PR #2, opened 2023-05-10
fix: Souce -> Source

ahoy-jon:
the smallest commit I made in my dev life
I would rather you fix it yourself and drop the PR!

_Also, it's a very interesting project, thank you for exploring this kind of design._


adamw:
:D Thanks!

PR #5, opened 2023-07-28
fix return type in race example in README

bishabosha:
it returns `2` yet the return type is `String`,


adamw:
Indeed, thanks :) We should move to mdoc soon ;)

PR #10, opened 2023-09-29
Fix useInScope example

efemelar:
This change 
 - adds import of `useInScope` that was missing
 - clarifies what value resource is acquired for 

Output before the change

```
acquiring ...
acquiring ...
Using MyResource(10) ...
Using MyResource(20) ...
releasing 20 ...
releasing 10 ...
```

and after change 

```
acquiring 10 ...
acquiring 20 ...
Using MyResource(10) ...
Using MyResource(20) ...
releasing 20 ...
releasing 10 ...
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -267,15 +267,16 @@ Resources can be allocated within a scope. They will be released in reverse acqu
 (that is, after all forks started within finish). E.g.:
 
 ```scala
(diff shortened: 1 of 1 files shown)


adamw:
Thanks! :)

PR #11, opened 2023-09-29
Fix supervised/scoped code example

efemelar:
Similar to https://github.com/softwaremill/ox/pull/10 fix for scoped useSupervised/useScoped


adamw:
Thanks! We should really migrate to mdoc soon ;)

PR #12, opened 2023-10-05
Add Source.interleave combinator

rucek:


core/src/test/scala/ox/channels/SourceOpsTest.scala
```diff
@@ -209,6 +209,51 @@ class SourceOpsTest extends AnyFlatSpec with Matchers with Eventually {
     }
   }
 
+  it should "interleave with an empty source" in scoped {
```
adamw, line 212: [minor] I would either move this to a separate test class (wont't `SourceOpsTest` turn into a monster-class?), or group these tests with `"interleave" should "xyz".

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -168,6 +168,67 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *     val s2 = Source.fromValues(10, 20, 30, 40)
+    *     s1.interleave(s2, segmentSize = 2).toList
+    *   }
+    *
+    *   scala> val res0: List[Int] = List(1, 2, 10, 20, 3, 4, 30, 40)
+    *   }}}
+    */
+  def interleave[U >: T](other: Source[U], segmentSize: Int = 1)(using Ox, StageCapacity): Source[U] =
+    val c = StageCapacity.newChannel[U]
+    var source: Source[U] = this
```
adamw, line 200: That's probably a cosmetic change, but maybe it would be clearer if we moved all the mutable data inside the fork, so that it's clear that it's the fork that "owns" the data, and hence mutable vars won't ever be accessed concurrently

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -168,6 +168,67 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    * @return
+    *   A source to which the interleaved elements from both sources would be sent.
+    * @example
+    *   {{{
+    *   scala>
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   scoped {
+    *     val s1 = Source.fromValues(1, 2, 3, 4)
```
adamw, line 190: maybe use uneven source sizes in the example

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -168,6 +168,67 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *     val s2 = Source.fromValues(10, 20, 30, 40)
+    *     s1.interleave(s2, segmentSize = 2).toList
+    *   }
+    *
+    *   scala> val res0: List[Int] = List(1, 2, 10, 20, 3, 4, 30, 40)
+    *   }}}
+    */
+  def interleave[U >: T](other: Source[U], segmentSize: Int = 1)(using Ox, StageCapacity): Source[U] =
```
adamw, line 198: to consider: optionally close the resulting source if either of the source-sources are done?

adamw:
Looks good, exactly the kind of PR I was looking for :)

core/src/main/scala/ox/channels/SourceOps.scala
rucek: Sure, good idea. 

Do we also want to add some more flexibility by allowing to interleave with more than one source (like https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/interleaveAll.html)? Then the internal implementation of `interleave` would be more generic, since at the moment it is "optimized" for exactly two sources.

core/src/test/scala/ox/channels/SourceOpsTest.scala
rucek: Done

core/src/main/scala/ox/channels/SourceOps.scala
rucek: Done

core/src/main/scala/ox/channels/SourceOps.scala
adamw: I guess we could simply add an `interleaveAll` method :)

PR #13, opened 2023-10-05
Add Source.interleaveAll combinator

rucek:


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -201,41 +202,86 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *     val s1 = Source.fromValues(1, 2, 3, 4, 5, 6, 7, 8)
+    *     val s2 = Source.fromValues(10, 20, 30)
+    *     val s3 = Source.fromValues(100, 200, 300, 400, 500)
+    *     s1.interleaveAll(List(s2, s3), segmentSize = 2, eagerComplete = true).toList
+    *   }
+    *
+    *   scala> val res0: List[Int] = List(1, 2, 10, 20, 100, 200, 3, 4, 30)
+    *   }}}
+    */
+  def interleaveAll[U >: T](others: Seq[Source[U]], segmentSize: Int = 1, eagerComplete: Boolean = false)(using
```
adamw, line 241: since none of the sources is special, maybe this shoul dbe a method on `Source` - `Source.interlaveAll`?

adamw: One downside is that this would require us to handle the case of an empty list - but then I think we just return a done source.

adamw (approved):
One suggestion, otherwise ok :)

core/src/main/scala/ox/channels/SourceOps.scala
rucek: Good point with moving it to `Source`. As for the empty list - we can either return a done source as you suggest, or change the signature of `interleaveAll` so that it enforces at least two sources (since one source would be a special case as well, where we just return the argument), e.g.
```scala
def interleaveAll[T](first: Source[T], second: Source[T], rest: Source[T]*)(segmentSize: Int, eagerComplete: Boolean)
```
The downside of the latter approach would be making the API inconsistent between `interleave` and `interleaveAll`. WDYT?

adamw: I don't think there's an inconsistency, `interleave` acts on two sources using the more convenient dot-notation, the `...All` variant takes an arbitrary list. The `first` / `second` approach would be good, but then it enforces static structure - won't work if you have a dynamically-defined list of sources.

PR #15, opened 2023-10-06
Add Source.mapStateful and Source.mapStatefulConcat combinators

rucek:
I'm not sure whether `f` should return `(S, Option[U])` or `(S, U)` - where in the former case the `Option` indicates whether anything should be sent downstream. 

The former is more flexible, but introduces additional complexity to the signature. Comparing to Akka Streams, the current approach mixes `statefulMap` and `statefulMapConcat` a bit, in that you can control whether the result of `f` will be sent downstream at all. However, unlike in Akka Streams, `f` can only return an `Option`, not an `Iterable` (which is flattened in Akka's `statefulMapConcat`).

Any thoughts on this (or anything else in the proposed approach)?

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -244,6 +244,62 @@ trait SourceOps[+T] { this: Source[T] =>
   def drain(): Unit = foreach(_ => ())
 
   def applied[U](f: Source[T] => U): U = f(this)
+
+  /** Applies the given mapping function `f`, using additional mutable state, to each element received from this source, and sends the
+    * results to the returned channel. Optionally sends an additional element, possibly based on the final state, to the returned channel
+    * once this source is done.
+    *
+    * The `initializeState` function is called once when `statefulMap` is called.
+    *
+    * The `onComplete` function is called once when this source is done. If it returns a non-empty value, the value will be sent to the
+    * returned channel, while an empty value will be ignored.
+    *
+    * @param initializeState
+    *   A function that initializes the state.
+    * @param f
+    *   A function that transforms the element from this source and the state into an optional pair of the next state and the result. If `f`
+    *   returns a nonempty value, th result will be sent to the returned channel, otherwise it will be ignored.
+    * @param onComplete
+    *   A function that transforms the final state into an optional element sent to the returned channel. By default the final state is
+    *   ignored.
+    * @return
+    *   A source to which the results of applying `f` to the elements from this source would be sent.
+    * @example
+    *   {{{
+    *   scala>
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   scoped {
+    *     val s = Source.fromValues(1, 2, 3, 4, 5)
+    *     s.statefulMap(() => 0)((sum, element) => (sum + element, Some(sum)), Some.apply)
+    *   }
+    *
+    *   scala> val res0: List[Int] = List(0, 1, 3, 6, 10, 15)
+    *   }}}
+    */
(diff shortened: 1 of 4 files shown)


rucek:
Also, any ideas for more test cases are welcome 😄

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -244,6 +244,62 @@ trait SourceOps[+T] { this: Source[T] =>
...
+  def statefulMap[S, U >: T](
+      initializeState: () => S
+  )(f: (S, T) => (S, Option[U]), onComplete: S => Option[U] = (_: S) => None)(using Ox, StageCapacity): Source[U] =
+    val c = StageCapacity.newChannel[U]
+    forkDaemon {
+      var state = initializeState()
+      repeatWhile {
+        receive() match
+          case ChannelClosed.Done =>
+            onComplete(state).foreach(c.send)
```
adamw, line 290: what if `onComplete` throws?

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -244,6 +244,62 @@ trait SourceOps[+T] { this: Source[T] =>
...
+            c.done()
+            false
+          case ChannelClosed.Error(r) =>
+            c.error(r)
+            false
+          case t: T @unchecked =>
+            val (nextState, result) = f(state, t)
```
adamw, line 297: same here - exception handling?

adamw:
I think that the akka streams approach might make sense - a general `mapStatefulConcat`, and a special-case, simplified `mapStateful`?

rucek:
See https://github.com/softwaremill/ox/pull/18#discussion_r1357039093 for a discussion about checking `isValue` after `send()`.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -244,6 +245,113 @@ trait SourceOps[+T] { this: Source[T] =>
   def drain(): Unit = foreach(_ => ())
 
   def applied[U](f: Source[T] => U): U = f(this)
+
+  /** Applies the given mapping function `f`, using additional mutable state, to each element received from this source, and sends the
```
adamw, line 249: small detail but might be important - I think I'd say `state`, not `mutable state`, as the `S` instance can (and probably often will be) immutable. There's nothing inherently mutable about `S` that is needed for this to work I guess?

rucek: True, fixed.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -244,6 +245,113 @@ trait SourceOps[+T] { this: Source[T] =>
...
+        receive() match
+          case ChannelClosed.Done =>
+            try
+              onComplete(state).foreach(c.send)
+              c.done()
+            catch case t: Throwable => c.error(t)
+            false
+          case ChannelClosed.Error(r) =>
+            c.error(r)
+            false
```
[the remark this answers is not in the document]
rucek: We're still checking the result of `c.send` below in `result.iterator.map(c.send).forall(_.isValue)`, although we assume that `send` should not fail since we control `c`, so we should probably always return `true` after sending, right?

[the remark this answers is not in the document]
adamw: right, though this at worst is dead code - not that there's a possiblity of incorrect behavior. We can change this though:
* ignoring the return value probably needs a comment that we know what we are doing - and why
* in larger code chunks (which are harder to follow), we might opt for the defensive solution anyway

PR #18, opened 2023-10-11
feat: introduce `intersperse` function

geminicaprograms:
The intersperse function behaves as `List.mkString` e.g.:
* called with an `inject` element only it produces

```scala
Source.fromValues("f", "b").intersperse(", ")  // ("f", ", ", "b")
```

* called with `start`, `inject` and `end` it produces

```scala
Source.fromValues("f", "b").intersperse("[", ", ", "]")  // ("[", "f", ", ", "b", "]")
```

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -2,7 +2,7 @@ package ox.channels
 
 import ox.*
 
-import java.util.concurrent.{ArrayBlockingQueue, ConcurrentLinkedQueue, CountDownLatch, LinkedBlockingQueue, Semaphore}
+import java.util.concurrent.{CountDownLatch, Semaphore}
 import scala.collection.mutable
 import scala.concurrent.duration.FiniteDuration
 
@@ -33,6 +33,67 @@ trait SourceOps[+T] { this: Source[T] =>
       }
     }
     c2
+
+  /** Intersperses this source with provided element and forwards it to the returned channel.
+    *
+    * @param inject
+    *   An element to be injected between the stream elements.
+    * @return
+    *   A source, onto which elements will be injected.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   scoped {
+    *     Source.empty[String].intersperse(", ").toList            // List()
+    *     Source.fromValues("foo").intersperse(", ").toList        // List(foo)
+    *     Source.fromValues("foo", "bar").intersperse(", ").toList // List(foo, ", ", bar)
+    *   }
+    *   }}}
+    */
+  def intersperse[U >: T](inject: U)(using Ox, StageCapacity): Source[U] =
+    intersperse(None, inject, None)
+
+  /** Intersperses this source with start, end and provided elements and forwards it to the returned channel.
+    *
+    * @param start
+    *   An element to be prepended to the stream.
+    * @param inject
+    *   An element to be injected between the stream elements.
+    * @param end
+    *   An element to be appended to the end of the stream.
+    * @return
+    *   A source, onto which elements will be injected.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   scoped {
+    *     Source.empty[String].intersperse("[", ", ", "]").toList            // List([, ])
+    *     Source.fromValues("foo").intersperse("[", ", ", "]").toList        // List([, foo, ])
+    *     Source.fromValues("foo", "bar").intersperse("[", ", ", "]").toList // List([, foo, ", ", bar, ])
+    *   }
+    *   }}}
+    */
+  def intersperse[U >: T](start: U, inject: U, end: U)(using Ox, StageCapacity): Source[U] =
+    intersperse(Some(start), inject, Some(end))
+
+  private def intersperse[U >: T](start: Option[U], inject: U, end: Option[U])(using Ox, StageCapacity): Source[U] =
+    val c = StageCapacity.newChannel[U]
+    forkDaemon {
+      start.foreach(c.send)
+      var firstEmitted = false
+      repeatWhile {
+        receive() match
+          case ChannelClosed.Done               => end.foreach(c.send); c.done(); false
+          case ChannelClosed.Error(e)           => c.error(e); false
+          case v: U @unchecked if !firstEmitted => firstEmitted = true; c.send(v); true
+          case v: U @unchecked                  => c.send(inject); c.send(v); true
+      }
+    }
+    c
 
   /** Applies the given mapping function `f` to each element received from this source, and sends the results to the returned channel. At
     * most `parallelism` invocations of `f` are run in parallel.
--- a/core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
+++ b/core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
@@ -1,0 +1,39 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsIntersperseTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.intersperse"
+
+  it should "intersperse with inject only over an empty source" in supervised {
+    val s = Source.empty[String]
+    s.intersperse(", ").toList shouldBe List.empty
+  }
+
+  it should "intersperse with inject only over a source with one element" in supervised {
+    val s = Source.fromValues("foo")
+    s.intersperse(", ").toList shouldBe List("foo")
+  }
+
+  it should "intersperse with inject only over a source with multiple elements" in supervised {
+    val s = Source.fromValues("foo", "bar")
+    s.intersperse(", ").toList shouldBe List("foo", ", ", "bar")
+  }
+
+  it should "intersperse with start, inject and end over an empty source" in supervised {
+    val s = Source.empty[String]
+    s.intersperse("[", ", ", "]").toList shouldBe List("[", "]")
+  }
+
+  it should "intersperse with start, inject and end over a source with one element" in supervised {
+    val s = Source.fromValues("foo")
+    s.intersperse("[", ", ", "]").toList shouldBe List("[", "foo", "]")
+  }
+
+  it should "intersperse with start, inject and end over a source with multiple elements" in supervised {
+    val s = Source.fromValues("foo", "bar")
+    s.intersperse("[", ", ", "]").toList shouldBe List("[", "foo", ", ", "bar", "]")
+  }
+}


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -34,6 +34,57 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    val c = StageCapacity.newChannel[T]
+    forkDaemon {
+      var firstEmitted = false
+      start.foreach(c.send(_))
+      repeatWhile {
+        receive() match
+          case ChannelClosed.Done               => end.foreach(c.send(_)); c.done(); false
+          case ChannelClosed.Error(e)           => c.error(e); false
+          case v: T @unchecked if !firstEmitted => firstEmitted = true; c.send(v).isValue
+          case v: T @unchecked                  => c.send(inject); c.send(v).isValue
```
geminicaprograms, line 83: **question:** `sbt compile` warns on this code fragment with
```
[warn] -- [E029] Pattern Match Exhaustivity Warning: /Users/jcentkowski/workspace/sml/ox/core/src/main/scala/ox/channels/SourceOps.scala:79:15 
[warn] 79 |        receive() match
[warn]    |        ^^^^^^^^^
[warn]    |        match may not be exhaustive.
[warn]    |
[warn]    |        It would fail on pattern case: _: T
[warn]    |
[warn]    | longer explanation available when compiling with `-explain`
```
but TBH I'm not sure when it would be actually `true` as both cases `ChannelClosed` (sealed trait with implementations covered) and value are handled 🤔  any suggestions on how to improve it are welcomed 🙇

geminicaprograms: The only solution that I can see to mute the warning is to add the following `case`
```scala
case _                                => c.error(new RuntimeException("Should never happen")); false
```
Would that make even sense?

kciesielski: @geminicaprograms I think it's suspicious that your intersperse methods are parameterized with `[T]`, as well as `SourceOps[+T]`. You're possibly dealing with two different `T`s here, which you don't want.

geminicaprograms: The problem was different but hint was great 🙇. The `T` cannot be used here as it would imply that covariant type can be used to form the resulting stream and as result existing (not necessarily covariant) elements would be treated as covariant type. One needs to use lower bound e.g. `U >: T` as parameter here so that existing elements are either of the same type or casted to their parent class.

Example with `T`:
```
trait Animal
class Dog extends Animal

val animals: Source[Animal] = Source[Animal]

animals.intersperse(new Dog) // the resulting stream is `Source[Dog]` that is potentially wrong for all other animals
```

Example with `U >: T`:
```
trait Animal
class Dog extends Animal

val animals: Source[Animal] = Source[Animal]

animals.intersperse(new Dog) // the resulting stream `Source[Animal]` that is always true
```

kciesielski (approved):

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -34,6 +34,57 @@ trait SourceOps[+T] { this: Source[T] =>
     }
     c2
 
+  /** Intersperses stream with provided element.
+    *
+    * @param inject
+    *   An element to be injected between the stream elements.
+    * @return
+    *   A source, onto which elements will be injected.
+    * @example
```
rucek, line 43: **suggestion:** Please add the necessary imports to the example (probably `import ox.*` and `import ox.channels.Source`) so that it's copy-paste'able (also in the other variant of `intersperse`)

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -34,6 +34,57 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *   }}}
+    */
+  def intersperse[U >: T](start: U, inject: U, end: U)(using Ox, StageCapacity): Source[U] =
+    intersperse(Some(start), inject, Some(end))
+
+  private def intersperse[U >: T](start: Option[U], inject: U, end: Option[U])(using Ox, StageCapacity): Source[U] =
+    val c = StageCapacity.newChannel[U]
+    forkDaemon {
+      var firstEmitted = false
+      start.foreach(c.send(_))
```
rucek, line 77: **nitpick:** Could be just `start.foreach(c.send)`, also applies to `case ChannelClosed.Done`

rucek: **question:** In the two other places we only continue processing when `send` succeeds, i.e. when `send(...).isValue` is `true` - shouldn't this be the case here as well?

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
```diff
@@ -0,0 +1,107 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsIntersperseTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.intersperse"
+
+  it should "intersperse with inject only over an empty source" in {
```
rucek, line 10: **nitpick:** You can avoid one level of the curly braces by writing
```scala
it should "..." in scoped {
  ...
}
```

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
```diff
@@ -0,0 +1,107 @@
...
+        c.send("bar")
+        c.done()
+      }
+
+      val s = c.intersperse(", ")
+
+      s.receive() shouldBe "foo"
+      s.receive() shouldBe ", "
+      s.receive() shouldBe "bar"
+      s.receive() shouldBe ChannelClosed.Done
```
rucek, line 52: **suggestion:** If you want to receive all elements from the source, you can use `s.toList` and make a single assertion on the list instead of aseerting individual `receive`s.

core/src/main/scala/ox/channels/SourceOps.scala
geminicaprograms: 🥇 Updated examples and confirmed in `Scala Worksheet` file

core/src/main/scala/ox/channels/SourceOps.scala
geminicaprograms: > Could be just start.foreach(c.send), also applies to case ChannelClosed.Done

👍 method reference rulez - fixed

geminicaprograms: > question: In the two other places we only continue processing...

That is necessary to properly implement the situation when sink cannot accept anything anymore, right? Fixed.

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
geminicaprograms: 💪  done

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
geminicaprograms: 💯  done

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -34,6 +34,68 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *   }
+    *   }}}
+    */
+  def intersperse[U >: T](start: U, inject: U, end: U)(using Ox, StageCapacity): Source[U] =
+    intersperse(Some(start), inject, Some(end))
+
+  private def intersperse[U >: T](start: Option[U], inject: U, end: Option[U])(using Ox, StageCapacity): Source[U] =
+    val c = StageCapacity.newChannel[U]
+    forkDaemon {
+      if (start.map(c.send(_).isValue).getOrElse(true))
```
adamw, line 86: we just created the `c` channel, it can't bet done/errored (we're the only ones sending to it)

adamw: @rucek yes, that's a good question, but I think it's the current design's assumption, that the streams are one-way - data always flows downstream. There's no backchannel which would allow propagating information. There are two ways of implementing such a backchannel, if needed:

(a) when the downstream receives an error/done, simply close the scope. This will close all intermediate components. My suspicion is that this will cover 90% of the use-cases, but I might be proven wrong of course. But so far I didn't come up with a compelling use-case to actually have the backchannel, in presence of structured concurrency & scopes
(b) for the other 10%, you can implement a backchannel using a channel. Then, in you original source or some intermediate component you can do `select(backchannel.receiveClause, s.sendClause(x))` to send data while being informed if anything is sent on the backchannel

The [docs](https://github.com/softwaremill/ox#error-propagation) kind of try to communicate the same as above, but maybe they need extending :)

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
```diff
@@ -0,0 +1,81 @@
...
+import ox.*
+
+class SourceOpsIntersperseTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.intersperse"
+
+  it should "intersperse with inject only over an empty source" in scoped {
+    val c = Channel[String]()
+    fork {
+      c.done()
+    }
```
rucek, line 14: **nitpick:** This is equivalent to `Source.empty[String]`

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
```diff
@@ -0,0 +1,81 @@
...
+
+    s.receive() shouldBe ChannelClosed.Done
+  }
+
+  it should "intersperse with inject only over a source with one element" in scoped {
+    val c = Channel[String]()
+    fork {
+      c.send("foo")
+      c.done()
+    }
```
rucek, line 26: **nitpick:** This is equivalent to `Source.fromValues("foo")`, you could also use `fromValues` in the other test cases.

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
geminicaprograms: Done

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
geminicaprograms: Done

core/src/main/scala/ox/channels/SourceOps.scala

rucek (approved):

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -34,6 +34,68 @@ trait SourceOps[+T] { this: Source[T] =>
...
+        var firstEmitted = false
+        repeatWhile {
+          receive() match
+            case ChannelClosed.Done               => end.foreach(c.send); c.done(); false
+            case ChannelClosed.Error(e)           => c.error(e); false
+            case v: U @unchecked if !firstEmitted => firstEmitted = true; c.send(v).isValue
```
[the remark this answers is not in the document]
geminicaprograms: done

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -34,6 +34,67 @@ trait SourceOps[+T] { this: Source[T] =>
     }
     c2
 
+  /** Intersperses stream with provided element.
```
geminicaprograms, line 37: **nitpick:** Updated function description to match the other operators

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -34,6 +34,67 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *     Source.empty[String].intersperse(", ").toList            // List()
+    *     Source.fromValues("foo").intersperse(", ").toList        // List(foo)
+    *     Source.fromValues("foo", "bar").intersperse(", ").toList // List(foo, ", ", bar)
+    *   }
+    *   }}}
+    */
+  def intersperse[U >: T](inject: U)(using Ox, StageCapacity): Source[U] =
+    intersperse(None, inject, None)
+
+  /** Intersperses stream with start, end and provided element.
```
geminicaprograms, line 58: **nitpick:** Updated function description to match the other operators

core/src/test/scala/ox/channels/SourceOpsIntersperseTest.scala
```diff
@@ -0,0 +1,39 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsIntersperseTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.intersperse"
+
+  it should "intersperse with inject only over an empty source" in scoped {
```
geminicaprograms, line 10: **suggestion:** use `supervised` instead of `scoped`

PR #19, opened 2023-10-16
feat: introduce `zipAll` function

geminicaprograms:
It combines elements from this and other sources in tuples handling ealy completion with defaults e.g.:

```scala
  Source.empty[Int].zipAll(Source.empty[String], -1, "foo").toList      // List()
  Source.empty[Int].zipAll(Source.fromValues("a"), -1, "foo").toList    // List((-1, "a"))
  Source.fromValues(1).zipAll(Source.empty[String], -1, "foo").toList   // List((1, "foo"))
  Source.fromValues(1).zipAll(Source.fromValues("a"), -1, "foo").toList // List((1, "a"))
```

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -2,7 +2,7 @@ package ox.channels
 
 import ox.*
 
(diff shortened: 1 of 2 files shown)


adamw:
Nice, thanks! :)

PR #20, opened 2023-10-16
feat: implement `drop` function

geminicaprograms:
Drop `n` elements from the source and emit subsequent elements (if any left) e.g.:

```scala
  Source.empty[Int].drop(1).toList          // List()
  Source.fromValues(1, 2, 3).drop(1).toList // List(2 ,3)
  Source.fromValues(1).drop(2).toList       // List()
```

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -2,7 +2,7 @@ package ox.channels
 
 import ox.*
 
-import java.util.concurrent.{ArrayBlockingQueue, ConcurrentLinkedQueue, CountDownLatch, LinkedBlockingQueue, Semaphore}
+import java.util.concurrent.{CountDownLatch, Semaphore}
 import scala.collection.mutable
 import scala.concurrent.duration.FiniteDuration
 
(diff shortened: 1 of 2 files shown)


core/src/test/scala/ox/channels/SourceOpsDropTest.scala
```diff
@@ -0,0 +1,24 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsDropTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.drop"
+
+  it should "not drop from the empty source" in scoped {
```
adamw, line 10: let's use `supervised` where possible; it should be our "default" scoping tool, unless some specific requirements necessitate usage of `scoped`

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -141,6 +141,24 @@ trait SourceOps[+T] { this: Source[T] =>
 
   def take(n: Int)(using Ox, StageCapacity): Source[T] = transform(_.take(n))
 
+  /** Drops `n` elements from the source and emits any subsequent elements.
```
adamw, line 144: I'd change the wording slightly to indicate that we return a different source than the current one. E.g.:

```
Drops `n` elements from this source and forwards any subsequent ones to the returned channel.
```

core/src/test/scala/ox/channels/SourceOpsDropTest.scala
```diff
@@ -0,0 +1,24 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsDropTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.drop"
```
rucek, line 8: **thought:** How about an additional test case to verify that `drop(0)` doesn't drop anything?

core/src/main/scala/ox/channels/SourceOps.scala

core/src/test/scala/ox/channels/SourceOpsDropTest.scala
geminicaprograms: Done

core/src/test/scala/ox/channels/SourceOpsDropTest.scala
geminicaprograms: Done

PR #21, opened 2023-10-16
feat: implement `dropWhile` function

geminicaprograms:
Drops elements from the source as long as the predicate is satisfied. Note that if the predicate fails then subsequent elements are no longer dropped even if they could still satisfy it.
Examples:

```scala
  Source.empty[Int].dropWhile(_ > 3).toList          // List()
  Source.fromValues(1, 2, 3).dropWhile(_ < 3).toList // List(3)
  Source.fromValues(1, 2, 1).dropWhile(_ < 2).toList // List(2, 1)
```

### DIFF
--- a/core/src/test/scala/ox/channels/SourceOpsDropWhileTest.scala
+++ b/core/src/test/scala/ox/channels/SourceOpsDropWhileTest.scala
@@ -1,0 +1,34 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsDropWhileTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.dropWhile"
+
+  it should "not drop from the empty source" in supervised {
+    val s = Source.empty[Int]
+    s.dropWhile(_ > 0).toList shouldBe List.empty
+  }
+
+  it should "drop elements from the source while predicate is true" in supervised {
+    val s = Source.fromValues(1, 2, 3)
+    s.dropWhile(_ < 3).toList shouldBe List(3)
+  }
+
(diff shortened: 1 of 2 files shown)


kciesielski (approved):

core/src/test/scala/ox/channels/SourceOpsDropWhileTest.scala
```diff
@@ -0,0 +1,34 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsDropWhileTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.drop"
```
kciesielski, line 8: **nitpick:** `dropWhile`

core/src/test/scala/ox/channels/SourceOpsDropWhileTest.scala
```diff
@@ -0,0 +1,34 @@
...
+
+  it should "not drop from the empty source" in scoped {
+    val s = Source.empty[Int]
+    s.dropWhile(_ > 0).toList shouldBe List.empty
+  }
+
+  it should "drop elements from the source while predicate is true" in scoped {
```
kciesielski, line 15: **suggestion:** If I understand correctly in Adam's review to another PR, `scoped` should be replaced with `supervised`

geminicaprograms: done

core/src/test/scala/ox/channels/SourceOpsDropWhileTest.scala
geminicaprograms: done

PR #22, opened 2023-10-16
feat: implement `takeWhile` function

geminicaprograms:
Takes elements from the source as long as the predicate is satisfied. Note that if the predicate fails then subsequent elements are not longer taken even if they could still satisfy it.
Example:
```scala
  Source.empty[Int].takeWhile(_ > 3).toList          // List()
  Source.fromValues(1, 2, 3).takeWhile(_ < 3).toList // List(1, 2)
  Source.fromValues(3, 2, 1).takeWhile(_ < 3).toList // List()
```

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -2,7 +2,7 @@ package ox.channels
 
 import ox.*
 
-import java.util.concurrent.{ArrayBlockingQueue, ConcurrentLinkedQueue, CountDownLatch, LinkedBlockingQueue, Semaphore}
+import java.util.concurrent.{CountDownLatch, Semaphore}
 import scala.collection.mutable
 import scala.concurrent.duration.FiniteDuration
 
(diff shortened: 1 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -141,6 +141,25 @@ trait SourceOps[+T] { this: Source[T] =>
 
   def take(n: Int)(using Ox, StageCapacity): Source[T] = transform(_.take(n))
 
+  /** Takes elements from the source as long as predicate `f` is satisfied (returns `true`). Note that once the predicate `f` is not
```
adamw, line 144: Same as in the others - I would change the wording a bit to indicate that the returned source is a separate one. E.g. `Forwards the elements from this source to the return one, as along ...`

rucek: We're using "sends to the returned channel" in other operators, so let's use a similar wording here for consistency from the user's perspective.

adamw:
Sorry, needs merge

adamw (approved):

geminicaprograms:
> Sorry, needs merge

Fixed 👍

PR #23, opened 2023-10-23
Don't rethrow exceptions in mapParUnordered

rucek:
Changes the error handling in `mapParUnordered` so that errors from upstream channel or from the mapping function are not rethrown, but only signalled to the downstream channel and used to terminate further processing. 

When the mapping function fails with an error, any already in-flight forks are allowed to complete, but their results are not sent downstream, since the downstream channel is closed immediately by the fork in error.

When the `receive()` from upstream fails, the downstream channel is closed and the internal loop terminates, so that no new forks are started.

Also adds `Source.failed(Throwable)` which was needed for one of the tests, so I decided not to put it in a separate PR.

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -188,8 +188,7 @@ trait SourceOps[+T] { this: Source[T] =>
                 try
                   c.send(f(t))
                   s.release()
-                catch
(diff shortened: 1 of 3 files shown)


adamw:
Great, thanks! :)

As a follow-up, can we do two things:
1. review existing operators if they don't throw exceptions in case of errors, but instead propagate them
2. update the channel error handling section of the readme to accurately describe our approach?

PR #24, opened 2023-10-24
feat: implement `throttle` function

geminicaprograms:
Sends elements to the returned channel limiting the throughput to specific number of elements (evenly spaced) per time unit. Note that the element's `receive()` time is included in the resulting throughput. For instance having `throttle(1, 1.second)` and `receive()` taking `Xms` means that resulting channel will receive elements every `1s + Xms` time. Throttling is not applied to the empty source.

Examples:
```scala
  Source.empty[Int].throttle(1, 1.second).toList       // List() returned without throttling
  Source.fromValues(1, 2).throttle(1, 1.second).toList // List(1, 2) returned after 2 seconds
```

Note that implementation relies on `Thread.sleep` which is (according to [What is blocking Loom?](https://softwaremill.com/what-is-blocking-in-loom/) project Loom compatible.

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -565,6 +565,46 @@ trait SourceOps[+T] { this: Source[T] =>
         case ChannelClosed.Error(r) => throw r.getOrElse(new NoSuchElementException("getting head failed"))
         case t: T @unchecked        => t
     }
+
+  /** Sends elements to the returned channel limiting the throughput to specific number of elements (evenly spaced) per time unit. Note that
+    * the element's `receive()` time is included in the resulting throughput. For instance having `throttle(1, 1.second)` and `receive()`
+    * taking `Xms` means that resulting channel will receive elements every `1s + Xms` time. Throttling is not applied to the empty source.
+    *
+    * @param elements
+    *   Number of elements to be emitted. Must be greater than 0.
+    * @param per
+    *   Per time unit. Must be greater or equal to 1 ms.
+    * @return
+    *   A source that emits at most `elements` `per` time unit.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   import scala.concurrent.duration.*
+    *
+    *   scoped {
+    *     Source.empty[Int].throttle(1, 1.second).toList       // List() returned without throttling
+    *     Source.fromValues(1, 2).throttle(1, 1.second).toList // List(1, 2) returned after 2 seconds
+    *   }
+    *   }}}
+    */
+  def throttle(elements: Int, per: FiniteDuration)(using Ox, StageCapacity): Source[T] =
+    require(elements > 0, "elements must be > 0")
+    require(per.toMillis > 0, "per time must be >= 1 ms")
+
+    val c = StageCapacity.newChannel[T]
+    val emitEveryMillis = per.toMillis / elements
+
+    forkDaemon {
+      repeatWhile {
+        receive() match
+          case ChannelClosed.Done     => c.done(); false
+          case ChannelClosed.Error(r) => c.error(r); false
+          case t: T @unchecked        => Thread.sleep(emitEveryMillis); c.send(t); true
+      }
+    }
+    c
 }
 
 trait SourceCompanionOps:
--- a/core/src/test/scala/ox/channels/SourceOpsThrottleTest.scala
+++ b/core/src/test/scala/ox/channels/SourceOpsThrottleTest.scala
@@ -1,0 +1,45 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+import scala.concurrent.duration.*
+
+class SourceOpsThrottleTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.throttle"
+
+  it should "not throttle the empty source" in supervised {
+    val s = Source.empty[Int]
+    val (result, executionTime) = measure { s.throttle(1, 1.second).toList }
+    result shouldBe List.empty
+    executionTime.toMillis should be < 1.second.toMillis
+  }
+
+  it should "throttle to specified elements per time units" in supervised {
+    val s = Source.fromValues(1, 2)
+    val (result, executionTime) = measure { s.throttle(1, 50.millis).toList }
+    result shouldBe List(1, 2)
+    executionTime.toMillis should (be >= 100L and be <= 150L)
+  }
+
+  it should "fail to throttle when elements <= 0" in supervised {
+    val s = Source.empty[Int]
+    the[IllegalArgumentException] thrownBy {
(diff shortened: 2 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -513,6 +514,48 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    require(elements > 0, "elements must be > 0")
+    require(per.toMillis > 0, "per time must be >= 1 ms")
+
+    val c = StageCapacity.newChannel[T]
+    val emitEveryMillis = per.toMillis / elements
+
+    forkDaemon {
+      val timer = new Timer
+      val task: TimerTask = () => {
+        receive() match
```
adamw, line 551: Hm doesn't this implement something slightly different? `Source.fromValues(1, 2).throttle(2, 1.second)` will complete in 500ms, instead of 0s (+/- couple of ms ;) ). Won't this be surprising?

geminicaprograms: `Source.fromValues(1, 2).throttle(2, 1.second)` means that element will be emitted every `500ms` and will complete just above `1 s` (as it will include also the emit time) and here is the test code for it:
```scala
    val (result, executionTime) = measure { s.throttle(2, 1.second).toList }
    result shouldBe List(1, 2)
    executionTime.toMillis should (be >= 1000L and be <= 1100L)
```
I was looking at this [definition](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/throttle.html#throttle) during the implementation 🤔

geminicaprograms: Here is another example
```scala
    Source
      .repeat(10)
      .throttle(1, 1.second)
      .foreach(n => println(s"${n} was emitted at ${LocalDateTime.now()}"))
```
which means that element will be emitted every `~1s` and here is the output:
```
10 was emitted at 2023-10-25T08:36:45.459262
10 was emitted at 2023-10-25T08:36:46.453793
10 was emitted at 2023-10-25T08:36:47.458780
10 was emitted at 2023-10-25T08:36:48.459819
...
```

core/src/test/scala/ox/channels/SourceOpsThrottleTest.scala
```diff
@@ -0,0 +1,46 @@
...
+  it should "throttle to specified elements per time units" in supervised {
+    val s = Source.fromValues(1, 2)
+    val (result, executionTime) = measure { s.throttle(1, 50.millis).toList }
+    result shouldBe List(1, 2)
+    executionTime.toMillis should (be >= 100L and be <= 150L)
+  }
+
+  it should "fail to throttle when elements <= 0" in supervised {
+    val s = Source.empty[Int]
+    the [IllegalArgumentException] thrownBy {
```
geminicaprograms: **nitpick (if-minor):** fix formatting in this line

core/src/main/scala/ox/channels/SourceOps.scala
adamw: Ok, let's update the docs then saying that the elements are evenly spaced within each window?

kciesielski: I'm not sure whether the even spacing is implied in the documentation of Akka Streams. Could you test it to make sure? Both Akka and [fs2](https://github.com/kovstas/fs2-throttler) solutions use additional notions of bursts, cost, and the token bucket algorithm, but maybe we don't want such a complex API in the first implementation.

geminicaprograms: Akka streams has 2 versions of `throttle`:
* simple one with the same arguments as in PR, it also works in the same fashion e.g.
  ```scala
   Source
    .repeat(10)
    .throttle(1, 1.second)
    .runWith(Sink.foreach(x => system.log.info(s"$x was emitted")))
  ```
  and that produces in the same manner:
  ```
  [INFO] [10/25/2023 12:06:20.398] [...] 10 was emitted
  [INFO] [10/25/2023 12:06:21.408] [...] 10 was emitted
  ...
  ```
* advanced when in addition to existing params `maximumBurst: Int` and  `mode: ThrottleMode` - I'm not sure if we need it now, but I'm happy to implement if that is correct

geminicaprograms: > Ok, let's update the docs then saying that the elements are evenly spaced within each window?

Added to the `throttle` description

kciesielski (approved):

kciesielski (approved):

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -513,6 +514,48 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    */
+  def throttle(elements: Int, per: FiniteDuration)(using Ox, StageCapacity): Source[T] =
+    require(elements > 0, "elements must be > 0")
+    require(per.toMillis > 0, "per time must be >= 1 ms")
+
+    val c = StageCapacity.newChannel[T]
+    val emitEveryMillis = per.toMillis / elements
+
+    forkDaemon {
+      val timer = new Timer
```
adamw, line 549: out of curiousity I looked up `Timer` and it seems it should not be used with Loom:
1. it uses `synchronized`, which is a no-go in loom (see "What's blocking in Loom")
2. it starts its own thread, which I think might end up being a platform thread, not a virtual one? The docs mention using `Object.wait`, instead of `Thread.sleep`, which for sure is Loom-friendly

Also, the javadocs mention that the task should be quick, otherwise:

```
Timer tasks should complete quickly. If a timer task takes excessive time to complete, it "hogs" the timer's task execution thread. This can, in turn, delay the execution of subsequent tasks, which may "bunch up" and execute in rapid succession when (and if) the offending task finally completes.
```

which I think might be a problem for us

geminicaprograms: 💪 I was wondering between `Thread.sleep`, `Object.wait` and `Timer` - mistakenly selected `Timer` (it has nicer API) - switched to `Thread.sleep` based implementation 🙇

core/src/test/scala/ox/channels/SourceOpsThrottleTest.scala
```diff
@@ -0,0 +1,46 @@
...
+    the[IllegalArgumentException] thrownBy {
+      s.throttle(1, 50.nanos)
+    } should have message "requirement failed: per time must be >= 1 ms"
+  }
+
+  private def measure[T](f: => T): (T, Duration) =
+    val before = System.currentTimeMillis()
+    val result = f
+    val after = System.currentTimeMillis();
+    (result, FiniteDuration.apply(after - before, TimeUnit.MILLISECONDS))
```
rucek, line 45: **nitpick:** Instead of using `FiniteDuration.apply`, you can use `(after - before).millis`

geminicaprograms: Done

rucek (approved):

PR #26, opened 2023-10-26
feat: implement `last` and `lastOption` operators

geminicaprograms:
The `lastOption` operator returns the last element in `Source` wrapped in `Some` or `None` in case when source is empty. Note that this is a terminal operation for source e.g.:

```scala
  Source.empty[Int].lastOption()  // None
  s.lastOption()                  // Some(2)
  s.receive()                     // ChannelClosed.Done
```

The `last` operator returns the last element in `Source` or throws `NoSuchElementException` in case when it is empty. In case when `receive()` fails with exception then this exception is re-thrown. It is also a terminal operation e.g.:

```scala
  Source.empty[Int].last()        // throws NoSuchElementException("cannot obtain last from an empty source")
  s.last()                        // 2
  s.receive()                     // ChannelClosed.Done
```

Note that `ChannelClosedException.Error` was improved to contain `cause` exception (if available).

### DIFF
--- a/core/src/main/scala/ox/channels/ChannelClosed.scala
+++ b/core/src/main/scala/ox/channels/ChannelClosed.scala
@@ -8,6 +8,6 @@ object ChannelClosed:
   case class Error(reason: Option[Throwable]) extends ChannelClosed
   case object Done extends ChannelClosed
 
-enum ChannelClosedException(reason: Option[Throwable]) extends Exception:
+enum ChannelClosedException(reason: Option[Throwable]) extends Exception(reason.orNull):
   case Error(reason: Option[Throwable]) extends ChannelClosedException(reason)
   case Done() extends ChannelClosedException(None)
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -605,6 +605,63 @@ trait SourceOps[+T] { this: Source[T] =>
       }
     }
     c
+
+  /** Returns the last element from this source wrapped in [[Some]] or [[None]] when this source is empty. Note that `lastOption` is a
+    * terminal operation leaving the source in [[ChannelClosed.Done]] state.
+    *
+    * @return
+    *   A `Some(last element)` if source is not empty or `None` otherwise.
+    * @throws ChannelClosedException.Error
+    *   When receiving an element from this source fails.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     Source.empty[Int].lastOption()  // None
+    *     val s = Source.fromValues(1, 2)
+    *     s.lastOption()                  // Some(2)
+    *     s.receive()                     // ChannelClosed.Done
+    *   }
+    *   }}}
+    */
+  def lastOption(): Option[T] =
+    supervised {
+      var value: Option[T] = None
+      repeatWhile {
+        receive() match
+          case ChannelClosed.Done     => false
+          case e: ChannelClosed.Error => throw e.toThrowable
+          case t: T @unchecked        => value = Some(t); true
+      }
+      value
+    }
+
+  /** Returns the last element from this source or throws [[NoSuchElementException]] when this source is empty. In case when receiving an
+    * element fails then [[ChannelClosedException.Error]] exception is thrown. Note that `last` is a terminal operation leaving the source
+    * in [[ChannelClosed.Done]] state.
+    *
+    * @return
+    *   A last element if source is not empty or throws otherwise.
+    * @throws NoSuchElementException
+    *   When source is empty.
+    * @throws ChannelClosedException.Error
+    *   When receiving an element from this source fails.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     Source.empty[Int].last()        // throws NoSuchElementException("cannot obtain last element from an empty source")
+    *     val s = Source.fromValues(1, 2)
+    *     s.last()                        // 2
+    *     s.receive()                     // ChannelClosed.Done
+    *   }
+    *   }}}
+    */
+  def last(): T = lastOption().getOrElse(throw new NoSuchElementException("cannot obtain last element from an empty source"))
 }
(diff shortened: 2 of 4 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,61 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    * Note that `lastOption` is a terminal operation leaving the source in `ChannelClosed.Done` state.
+    *
+    * @return
+    *   A `Some(last element)` if source is not empty or None` otherwise.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   scoped {
```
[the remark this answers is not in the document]

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,61 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *   }
+    *   }}}
+    */
+  def last(): T =
+    supervised {
+      var value: Option[T] = None
+      repeatUntil {
+        receive() match
+          case ChannelClosed.Done     => true
+          case ChannelClosed.Error(r) => throw r.getOrElse(new NoSuchElementException("getting last failed"))
```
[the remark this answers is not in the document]

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,61 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *
+    *   supervised {
+    *     Source.empty[Int].lastOption()  // None
+    *     val s = Source.fromValues(1, 2)
+    *     s.lastOption()                  // Some(2)
+    *     s.receive()                     // ChannelClosed.Done
+    *   }
+    *   }}}
+    */
+  def lastOption(): Option[T] = Try(last()).toOption
```
adamw, line 587: shouldn't `lastOption` throw an exception, if there's an error, and only return `None` on empty (done) sources? That way we suppress the exception, and that's never a good thing

also, then we can implement `last` in terms of `lastOption`, not the other way round, and avoid throwing-catching

geminicaprograms: done

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,63 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *     val s = Source.fromValues(1, 2)
+    *     s.lastOption()                  // Some(2)
+    *     s.receive()                     // ChannelClosed.Done
+    *   }
+    *   }}}
+    */
+  def lastOption(): Option[T] =
+    supervised {
+      var value: Option[T] = None
+      repeatUntil {
```
rucek, line 592: **nitpick:** In other operators we're using `repeatWhile`, which is similar, just with the condition inverted. I think we could use `repeatWhile` here as well for consistency.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,63 @@ trait SourceOps[+T] { this: Source[T] =>
         case ChannelClosed.Error(r) => throw r.getOrElse(new NoSuchElementException("getting head failed"))
         case t: T @unchecked        => t
     }
+
+  /** Returns the last element from this source wrapped in [[Some]] or [[None]] when the source is empty. Note that `lastOption` is a
```
rucek, line 569: **suggestion:** According to the [scaladoc style guide](https://docs.scala-lang.org/style/scaladoc.html#general-style), please use "this source" instead of "[the] source" throughout the scaladoc. Also applies to the scaladoc for `last`.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,63 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    * terminal operation leaving the source in [[ChannelClosed.Done]] state.
+    *
+    * @return
+    *   A `Some(last element)` if source is not empty or `None` otherwise.
+    * @throws ChannelClosedException.Error
+    *   When [[receive]] fails.
```
rucek, line 575: **thought:** `receive` (the function) is an implementation detail that the user might not be aware of. How about rephrasing this like "When receiving an element from this source fails."? Also applies to `last`

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,63 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *
+    *   supervised {
+    *     Source.empty[Int].last()        // throws NoSuchElementException("cannot obtain last from an empty source")
+    *     val s = Source.fromValues(1, 2)
+    *     s.last()                        // 2
+    *     s.receive()                     // ChannelClosed.Done
+    *   }
+    *   }}}
+    */
+  def last(): T = lastOption().getOrElse(throw new NoSuchElementException("cannot obtain last from an empty source"))
```
rucek, line 624: **nitpick:** Please use "last element" insted of "last" in the error message.

core/src/main/scala/ox/channels/ChannelClosed.scala
```diff
@@ -8,6 +8,6 @@ object ChannelClosed:
   case class Error(reason: Option[Throwable]) extends ChannelClosed
   case object Done extends ChannelClosed
 
-enum ChannelClosedException(reason: Option[Throwable]) extends Exception:
+enum ChannelClosedException(reason: Option[Throwable]) extends Exception(reason.orNull):
```
rucek, line 11: **praise:** Nice catch

adamw (approved):

core/src/main/scala/ox/channels/SourceOps.scala

core/src/main/scala/ox/channels/SourceOps.scala

core/src/main/scala/ox/channels/SourceOps.scala
geminicaprograms: 🤷‍♂️  but turned to `repeatWhile`

core/src/main/scala/ox/channels/SourceOps.scala

rucek (approved):

PR #27, opened 2023-10-26
feat: implement `fold` operator

geminicaprograms:
The `fold` operation returns combined value retrieved from running function `f` on all source elements in a cumulative manner where result of the previous call is used as an input value to the next e.g.:

```scala
  Source.empty[Int].fold(0)((acc, n) => acc + n)       // 0
  Source.fromValues(2, 3).fold(5)((acc, n) => acc - n) // 0
```

Note that in case when `receive()` operation fails then:
* the original exception is re-thrown
* `NoSuchElement` exception is thrown when source fails without error
* when function `f` fails with exception then this exception is propagated up to the caller

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -46,7 +46,7 @@ trait SourceOps[+T] { this: Source[T] =>
     *   import ox.*
     *   import ox.channels.Source
     *
-    *   scoped {
+    *   supervised {
     *     Source.empty[String].intersperse(", ").toList            // List()
     *     Source.fromValues("foo").intersperse(", ").toList        // List(foo)
     *     Source.fromValues("foo", "bar").intersperse(", ").toList // List(foo, ", ", bar)
@@ -71,7 +71,7 @@ trait SourceOps[+T] { this: Source[T] =>
     *   import ox.*
     *   import ox.channels.Source
     *
-    *   scoped {
+    *   supervised {
     *     Source.empty[String].intersperse("[", ", ", "]").toList            // List([, ])
     *     Source.fromValues("foo").intersperse("[", ", ", "]").toList        // List([, foo, ])
     *     Source.fromValues("foo", "bar").intersperse("[", ", ", "]").toList // List([, foo, ", ", bar, ])
@@ -210,7 +210,7 @@ trait SourceOps[+T] { this: Source[T] =>
     *   import ox.*
     *   import ox.channels.Source
     *
-    *   scoped {
+    *   supervised {
     *     Source.empty[Int].takeWhile(_ > 3).toList          // List()
     *     Source.fromValues(1, 2, 3).takeWhile(_ < 3).toList // List(1, 2)
     *     Source.fromValues(3, 2, 1).takeWhile(_ < 3).toList // List()
@@ -228,7 +228,7 @@ trait SourceOps[+T] { this: Source[T] =>
     *   import ox.*
     *   import ox.channels.Source
     *
-    *   scoped {
+    *   supervised {
     *     Source.empty[Int].drop(1).toList          // List()
     *     Source.fromValues(1, 2, 3).drop(1).toList // List(2 ,3)
     *     Source.fromValues(1).drop(2).toList       // List()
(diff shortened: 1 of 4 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,44 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    * @throws NoSuchElementException
+    *   When `receive()` failed without error.
+    * @throws exception
+    *   When `receive()` failed with exception then this exception is re-thrown.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   scoped {
```
adamw, line 589: supervised :) scoped is more of a low-level construct

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -565,6 +565,44 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *   }
+    *   }}}
+    */
+  def fold[U](zero: U)(f: (U, T) => U): U =
+    supervised {
+      var current = zero
+      repeatWhile {
+        receive() match
+          case ChannelClosed.Done     => false
+          case ChannelClosed.Error(r) => throw r.getOrElse(new NoSuchElementException("folding failed"))
```
adamw, line 601: you can call ChannelClosed.toThrowable

core/src/main/scala/ox/channels/SourceOps.scala

core/src/main/scala/ox/channels/SourceOps.scala

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -623,6 +623,42 @@ trait SourceOps[+T] { this: Source[T] =>
     *   }}}
     */
   def last(): T = lastOption().getOrElse(throw new NoSuchElementException("cannot obtain last from an empty source"))
+
+  /** Uses `zero` as the current value and applies function `f` on it and a value received from a source. The returned value is used as the
```
rucek, line 627: **suggestion:** Please see the scaladoc-related comments for `last`/`lastOption`:
- https://github.com/softwaremill/ox/pull/26#discussion_r1374975898
- https://github.com/softwaremill/ox/pull/26#discussion_r1374977865

core/src/test/scala/ox/channels/SourceOpsFoldTest.scala
```diff
@@ -0,0 +1,39 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsFoldTest extends AnyFlatSpec with Matchers {
```
rucek, line 7: **thought:** Can we have another test case for a scenario where the function passed to `fold` throws an exception?

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -623,6 +623,42 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *   }}}
+    */
+  def fold[U](zero: U)(f: (U, T) => U): U =
+    supervised {
+      var current = zero
+      repeatWhile {
+        receive() match
+          case ChannelClosed.Done     => false
+          case e: ChannelClosed.Error => throw e.toThrowable
+          case t: T @unchecked        => current = f(current, t); true
```
adamw, line 658: this lacks error handling, if `f` would throw an exception

adamw: I take that back, if `f` throws an exception, it simply propagates upwards, nothing to do here :)

core/src/main/scala/ox/channels/SourceOps.scala

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -623,6 +623,42 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     Source.empty[Int].fold(0)((acc, n) => acc + n)       // 0
+    *     Source.fromValues(2, 3).fold(5)((acc, n) => acc - n) // 0
+    *   }
+    *   }}}
+    */
+  def fold[U](zero: U)(f: (U, T) => U): U =
+    supervised {
```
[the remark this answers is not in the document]
geminicaprograms: we don't need it - thanks!

core/src/main/scala/ox/channels/SourceOps.scala
geminicaprograms: 👍 - added note to scaladoc about this behaviour

core/src/test/scala/ox/channels/SourceOpsFoldTest.scala
geminicaprograms: Done

adamw (approved):

rucek (approved):

PR #29, opened 2023-10-27
Unify error handling

rucek:
- `mapParScope` doesn't rethrow anymore
- `pipeTo` doesn't rethrow when receiving from upstream fails
- updated docs on error propagation, with an example based on `mapParUnordered`

### DIFF
--- a/README.md
+++ b/README.md
@@ -628,10 +628,76 @@ exception being thrown there.
 Won't this design cause upstream channels / sources to operate despite the consumer being gone (because of the 
 exception)?
 
-No: the exception should cause the containing scope to finish (or a supervised fork to fail), cancelling any forks 
-that are operating in the background. Any unused channels can then be garbage-collected.
+It depends on two factors:
+- whether there are any forks running in parallel to the failed one,
+- whether you only signal the exception downstream, or also choose to re-throw it.
 
-The role of the exception handler is then to re-create the entire processing pipeline, or escalate the error further.
+If there's only a single fork running at a time, it would terminate processing anyway, so it's enough to signal the exception to the downstream.
+
+If there are multiple forks running in parallel, there are two possible scenarios:
+1. If you choose to re-throw the exception, it should cause the containing scope to finish (or a supervised fork to fail), 
+cancelling any forks that are operating in the background. Any unused channels can then be garbage-collected.
+2. If you choose not to re-throw, the forks running in parallel would be allowed to complete normally (unless the containing scope is finished for another reason).
+
+Internally, for the built-in `Source` operators, we took the latter approach, i.e. we chose not to re-throw and let the parrallel forks complete normally. 
+However, we keep in mind that they might not be able to send to downstream channel anymore - since the downstream might already be closed by the failing fork.
+
+### Example
+
+Let's have a look at the error handling in `Source.mapParUnordered` to demonstrate our approach. This operator applies a mapping function to a given number of elements in parallel, and is implemented as follows:
+
+```scala
+def mapParUnordered[U](parallelism: Int)(f: T => U)(using Ox, StageCapacity): Source[U] =
+  val c = StageCapacity.newChannel[U]
+  val s = new Semaphore(parallelism)
+  forkDaemon {
+    supervised {                                          // (1)
+      repeatWhile {                                       
+        s.acquire()
+        receive() match
+          case ChannelClosed.Done => false
+          case e @ ChannelClosed.Error(r) =>              // (2)
+            c.error(r)
+            false
+          case t: T @unchecked =>
+            fork {                                        // (3)
+              try
+                c.send(f(t))                              // (4)
+                s.release()
+              catch case t: Throwable => c.error(t)       // (5)
+            }
+            true
+      }
+    }
+    c.done()
+  }
+  c
+```
+
+It first creates a `supervised` scope (1), i.e. one that only completes (on the happy path) when all 
+non-daemon supervised forks complete. The mapping function `f` is then run in parallel using non-daemon `fork`s (3).
+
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -628,10 +628,76 @@ exception being thrown there.
 Won't this design cause upstream channels / sources to operate despite the consumer being gone (because of the 
 exception)?
 
-No: the exception should cause the containing scope to finish (or a supervised fork to fail), cancelling any forks 
-that are operating in the background. Any unused channels can then be garbage-collected.
+It depends on two factors:
```
adamw, line 631: I think I would mention beforehand that there are two possible designs, of which we picked one, as otherwise this might make the impression that error handling in Ox is dependent on the factors below (which it isn't)

adamw: In fact ... such a discussion would be better placed in and ADR. Maybe this should be moved there? ADRs can be as simple as a text file, e.g. here: https://github.com/softwaremill/tapir/tree/master/doc/adr

The docs can then simply state the current state of affairs, keeping the design discussions separate

rucek: Good point, moved the discussion to an ADR.

doc/adr/0001-error-propagation-in-channels.md
```diff
@@ -0,0 +1,87 @@
...
+What should happen when an error is encountered when processing channel elements? Should it be propagated downstream or re-thrown?
+
+## Decision
+
+We chose to only propagate the errors downstream, so that they are eventually thrown when the source is discharged.
+
+Won't this design cause upstream channels / sources to operate despite the consumer being gone (because of the 
+exception)?
+
+It might if there are mutliple forks running in parallel, of which one end with an error. This
```
geminicaprograms, line 16: **nitpick (if-minor):** there is orphaned ` This` in the of sentence

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -157,8 +157,9 @@ trait SourceOps[+T] { this: Source[T] =>
       fork {
         repeatWhile {
           inProgress.receive() match
-            case f: Fork[U] @unchecked =>
-              c2.send(f.join()).isValue
+            case f: Fork[Option[U]] @unchecked =>
+              f.join().foreach(c2.send)
```
geminicaprograms, line 161: **question:** Should we keep sending when `Option[U]` is `None`? At that point `c2` received already an `error` so how about returning `false` in such case?

doc/adr/0001-error-propagation-in-channels.md
rucek: Good catch, thx, actually this entire paragraph was maant to be temporary

doc/adr/0001-error-propagation-in-channels.md
```diff
@@ -0,0 +1,87 @@
...
+def mapParUnordered[U](parallelism: Int)(f: T => U)(using Ox, StageCapacity): Source[U] =
+  val c = StageCapacity.newChannel[U]
+  val s = new Semaphore(parallelism)
+  forkDaemon {
+    supervised {                                          // (1)
+      repeatWhile {                                       
+        s.acquire()
+        receive() match
+          case ChannelClosed.Done => false
+          case e @ ChannelClosed.Error(r) =>              // (2)
```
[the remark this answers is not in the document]
rucek: Right, thx, also fixed this at the source in `SourceOps`.

doc/adr/0001-error-propagation-in-channels.md
```diff
@@ -0,0 +1,87 @@
...
+val input: Source[Int] = Source.fromValues(1, 2, 3, 4)
+def f(i: Int): Int = if ()
+
+val result: Source[Int] = input.mapParUnordered(2)(f)
+```
+
+Let's also assume that the mapping function `f` is an identity with a fixed delay, but it's going to fail 
+immediately (by throwing an exception) when it processes the third element.
+
+In this scenario, the first 2-element batch would successfully process elements `1` and `2`, and emit them
```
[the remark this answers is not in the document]
rucek: I'd argue that mine is correct, since it's an adjective, see e.g. https://ell.stackexchange.com/a/5580. Similarly to a "2-hour drive", a "two-day streak" or a "5-minute walk".

core/src/main/scala/ox/channels/SourceOps.scala
rucek: You're right, fixed.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -157,8 +157,8 @@ trait SourceOps[+T] { this: Source[T] =>
       fork {
         repeatWhile {
           inProgress.receive() match
-            case f: Fork[U] @unchecked =>
-              c2.send(f.join()).isValue
+            case f: Fork[Option[U]] @unchecked =>
+              f.join().map(c2.send).isDefined
```
geminicaprograms, line 161: **question:**  What if `exists` was used instead:
```scala
f.join().exists(c2.send(_).isValue)
```
AFAIU these two cases would be covered:
1. `None` was sent hence there is no point in sending more elements
2. `Some(x)` was sent but destination channel failed to receive it, again there is no point in sending more elements

🤔 WDYT?

geminicaprograms (approved):

core/src/main/scala/ox/channels/SourceOps.scala
rucek: `c2` is a channel that we own/control (it's created internally in `mapPar`). IIRC we agreed that `send`ing to it can't fail due to that, and that we don't need to verify `isValue` after sending. Is that correct or am I missing something?

adamw (approved):

PR #30, opened 2023-10-27
feat: implement `reduce` operator

geminicaprograms:
Uses the first and the following (if available) elements from this source and applies function `f` on them. The returned value is used as the next current value and `f` is applied again with the value received from a source. The operation is repeated until the source is drained. This is similar operation to `fold` but it uses the first source element as `zero` e.g.:

```scala
  Source.empty[Int].reduce(_ + _)    // throws NoSuchElementException("cannot reduce an empty source")
  Source.fromValues(1).reduce(_ + _) // 1
  s.reduce(_ + _)                    // 3
  s.receive()                        // ChannelClosed.Done
```

Note that when function `f` fails then exception is propagated up to the caller.

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -699,6 +699,38 @@ trait SourceOps[+T] { this: Source[T] =>
         case t: T @unchecked        => current = f(current, t); true
     }
     current
+
+  /** Uses the first and the following (if available) elements from this source and applies function `f` on them. The returned value is used
+    * as the next current value and `f` is applied again with the value received from this source. The operation is repeated until this
+    * source is drained. This is similar operation to [[fold]] but it uses the first source element as `zero`.
+    *
+    * @param f
+    *   A binary function (a function that takes two arguments) that is applied to the current and next values received from this source.
+    * @return
+    *   Combined value retrieved from running function `f` on all source elements in a cumulative manner where result of the previous call
+    *   is used as an input value to the next.
+    * @throws NoSuchElementException
+    *   When this source is empty.
+    * @throws ChannelClosedException.Error
+    *   When receiving an element from this source fails.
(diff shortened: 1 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -659,6 +659,36 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *     Source.empty[Int].reduce(_ + _)    // throws NoSuchElementException("cannot reduce over an empty source")
+    *     Source.fromValues(1).reduce(_ + _) // 1
+    *     val s = Source.fromValues(1, 2)
+    *     s.reduce(_ + _)                    // 3
+    *     s.receive()                        // ChannelClosed.Done
+    *   }
+    *   }}}
+    */
+  def reduce[U >: T](f: (U, U) => U): U =
+    fold(headOption().getOrElse(throw new NoSuchElementException("cannot reduce over an empty source")))(f)
```
rucek, line 691: **nitpick:** (wording) I think we reduce "a source", not "over a source", since it's about reducing a collection to a single element (also applies to the example in scaladoc).

adamw:
I fixed the title ;)

core/src/main/scala/ox/channels/SourceOps.scala

core/src/test/scala/ox/channels/SourceOpsFoldTest.scala
```diff
@@ -0,0 +1,39 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+class SourceOpsFoldTest extends AnyFlatSpec with Matchers {
```
geminicaprograms: **issue:** add test when exception is thrown in function `f`

rucek (approved):

PR #31, opened 2023-10-30
feat: introduce `takeLast(n)` operator

geminicaprograms:
Returns the list of up to `n` last elements from this source. Less than `n` elements is returned when this source contains les elements than requested. The [[List.empty]] is returned when `takeLast` is called on an empty source.

Example:
```scala
  Source.empty[Int].takeLast(5)    // List.empty
  Source.fromValues(1).takeLast(0) // List.empty
  Source.fromValues(1).takeLast(2) // List(1)
  val s = Source.fromValues(1, 2, 3, 4)
  s.takeLast(2)                    // List(4, 5)
  s.receive()                      // ChannelClosed.Done
```

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -2,6 +2,7 @@ package ox.channels
 
 import ox.*
 
+import java.util
 import java.util.concurrent.{CountDownLatch, Semaphore}
 import scala.collection.{IterableOnce, mutable}
 import scala.concurrent.duration.FiniteDuration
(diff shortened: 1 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -731,6 +732,45 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *     Source.empty[Int].takeLast(5)    // List.empty
+    *     Source.fromValues(1).takeLast(2) // List(1)
+    *     val s = Source.fromValues(1, 2, 3, 4)
+    *     s.takeLast(2)                    // List(4, 5)
+    *     s.receive()                      // ChannelClosed.Done
+    *   }
+    *   }}}
+    */
+  def takeLast(n: Int): List[T] =
+    require(n > 0, "n must be > 0")
```
rucek, line 760: **thought:** Isn't `0` a legal case as well, where we would return an empty list? We'd probably need to `drain` the source then to make the side effects "compatible" with calling `takeLast` with `n` > 0.

kciesielski: Akka Streams [requires n > 0](https://github.com/akka/akka/blob/56bedb33b34523ccb8536b8a989f0a76d2865d81/akka-stream/src/main/scala/akka/stream/impl/Sinks.scala#L151). On the other hand, I don't see anything wrong with your idea @rucek.

rucek: OK, that was probably the inspiration. On the other hand, Scala collections use `n.max(0)` - although I'm aware that collections might have different semantics than streams/channels. My own issue with allowing `0` for `takeLast` is that it might require "artificially" draining the channel - I'm not sure if it's a correct approach.

@adamw any thoughts?

adamw: Nil + drain sounds most reasonable I guess :)

geminicaprograms: Added the case when `n == 0` to implementation and unit tests.

adamw (approved):

PR #32, opened 2023-10-31
feat: implement `Source.future` operator

geminicaprograms:
Creates a source that emits a single value when Future completes or fails otherwise.

Notes:
* when Future fails with `scala.concurrent.ExecutionException` then its cause is returned as source failure.
* the input Future cannot be `null`
* the `Future` completion is performed on the provided `ExecutionContext`

Examples:
```scala
  Source .future(Future.failed(new RuntimeException("future failed")))
    .receive()                               // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
  Source.future(Future.successful(1)).toList // List(1)
```

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -6,6 +6,8 @@ import java.util
 import java.util.concurrent.{CountDownLatch, Semaphore}
 import scala.collection.{IterableOnce, mutable}
 import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.{ExecutionContext, ExecutionException, Future}
+import scala.util.{Failure, Success}
 
 trait SourceOps[+T] { this: Source[T] =>
   // view ops (lazy)
@@ -992,6 +994,39 @@ trait SourceCompanionOps:
           }
         }
         c
+
+  /** Creates a source that emits a single value when `from` completes or fails otherwise. The `from` completion is performed on the
+    * provided [[scala.concurrent.ExecutionContext]]. Note that when `from` fails with [[scala.concurrent.ExecutionException]] then its
+    * cause is returned as source failure.
+    *
+    * @param from
+    *   A [[scala.concurrent.Future]] that returns value upon completion.
+    * @return
+    *   A source that will emit value upon a `from` [[scala.concurrent.Future]] completion.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   import scala.concurrent.ExecutionContext.Implicits.global
+    *   import scala.concurrent.Future
+    *
+    *   supervised {
+    *     Source
+    *       .future(Future.failed(new RuntimeException("future failed")))
+    *       .receive()                               // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
+    *     Source.future(Future.successful(1)).toList // List(1)
+    *   }
+    *   }}}
+    */
+  def future[T](from: Future[T])(using StageCapacity, ExecutionContext): Source[T] =
+    val c = StageCapacity.newChannel[T]
+    from.onComplete {
+      case Success(value)                  => c.send(value); c.done()
+      case Failure(ex: ExecutionException) => c.error(ex.getCause)
+      case Failure(ex)                     => c.error(ex)
+    }
+    c
 
   /** Creates a source that fails immediately with the given [[java.lang.Throwable]]
     *
--- a/core/src/test/scala/ox/channels/SourceOpsFutureTest.scala
+++ b/core/src/test/scala/ox/channels/SourceOpsFutureTest.scala
@@ -1,0 +1,29 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+
+import scala.concurrent.Future
+
+class SourceOpsFutureTest extends AnyFlatSpec with Matchers {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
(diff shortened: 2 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -993,6 +994,42 @@ trait SourceCompanionOps:
...
+    *     Source.future(Future.successful(1)).toList // List(1)
+    *   }
+    *   }}}
+    */
+  def future[T](from: Future[T])(using Ox, StageCapacity): Source[T] =
+    require(from != null, "from cannot be null")
+    val c = StageCapacity.newChannel[T]
+    forkDaemon {
+      try
+        val t = Await.result(from, Duration.Inf)
```
geminicaprograms, line 1024: **question:** ATM `Duration.Inf` is used but it could be promoted to the default value of `timeout` parameter.
I was also considering going down to `Java` with `val t = FutureConverters.FutureOps(from).asJava.toCompletableFuture.get()` but:
* it seems to be needed as AFAIU `Await.result` is Loom safe
* we could stay in `Scala` land

comments are welcomed 💪

core/src/test/scala/ox/channels/SourceOpsFutureTest.scala
```diff
@@ -0,0 +1,26 @@
...
+class SourceOpsFutureTest extends AnyFlatSpec with Matchers {
+  behavior of "Source.future"
+
+  it should "fail to create source from the null future" in supervised {
+    the[IllegalArgumentException] thrownBy {
+      Source.future(null)
+    } should have message "requirement failed: from cannot be null"
+  }
+
+  it should "return report failure when future fails" in supervised {
```
geminicaprograms, line 18: **nitpick:** `return report` looks like typo

geminicaprograms: Done

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -993,6 +995,40 @@ trait SourceCompanionOps:
...
+    * @param from
+    *   A [[scala.concurrent.Future]] that returns value upon completion. It cannot be `null`.
+    * @return
+    *   A source that will emit value upon a `from` [[scala.concurrent.Future]] completion.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   import scala.concurrent.Future
```
geminicaprograms, line 1010: **issue:** missing `import scala.concurrent.ExecutionContext.Implicits.global`

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -993,6 +995,40 @@ trait SourceCompanionOps:
...
+    *       .future(Future.failed(new RuntimeException("future failed")))
+    *       .receive()                               // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
+    *     Source.future(Future.successful(1)).toList // List(1)
+    *   }
+    *   }}}
+    */
+  def future[T](from: Future[T])(using Ox, StageCapacity, ExecutionContext): Source[T] =
+    require(from != null, "from cannot be null")
+    val c = StageCapacity.newChannel[T]
+    forkDaemon {
```
geminicaprograms, line 1023: **issue:** we can safely rely on `ec` (and caller's decision) on which thread send should be performed

core/src/main/scala/ox/channels/SourceOps.scala
geminicaprograms: done

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -993,6 +995,39 @@ trait SourceCompanionOps:
         }
         c
 
+  /** Creates a source that emits a single value when `from` completes or fails otherwise. Note that when `from` fails with
+    * [[scala.concurrent.ExecutionException]] then its cause is returned as source failure.
```
geminicaprograms, line 999: Add information that `Future` completion is performed on the provided `ExecutionContext`

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -993,6 +995,40 @@ trait SourceCompanionOps:
...
+    *     Source.future(Future.successful(1)).toList // List(1)
+    *   }
+    *   }}}
+    */
+  def future[T](from: Future[T])(using StageCapacity, ExecutionContext): Source[T] =
+    require(from != null, "from cannot be null")
+    val c = StageCapacity.newChannel[T]
+    from.onComplete {
+      case Success(value)                => c.send(value); c.done()
+      case Failure(ex: ExecutionContext) => c.error(ex.getCause)
```
kciesielski, line 1027: **issue** `ExecutionContext` is not a `Throwable`, I wonder how come this compiles at all :thinking:

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -993,6 +995,40 @@ trait SourceCompanionOps:
...
+    *   supervised {
+    *     Source
+    *       .future(Future.failed(new RuntimeException("future failed")))
+    *       .receive()                               // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
+    *     Source.future(Future.successful(1)).toList // List(1)
+    *   }
+    *   }}}
+    */
+  def future[T](from: Future[T])(using StageCapacity, ExecutionContext): Source[T] =
+    require(from != null, "from cannot be null")
```
adamw, line 1023: do we want this kind of defences? `null`s are essentially unused in Scala

rucek (approved):

PR #33, opened 2023-10-31
feat: implement `Source.futureSource` operator

geminicaprograms:
Creates a source that emits elements from future source when it completes or fails otherwise. The future completion is performed on the provided `ExecutionContext` whereas elements are emitted through `Supervised`. Note that when Future fails with `ExecutionException` then its cause is returned as source failure.

Examples:
```scala
  Source
    .futureSource(Future.failed(new RuntimeException("future failed")))
    .receive()                                                           // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
  Source.futureSource(Future.successful(Source.fromValues(1, 2))).toList // List(1, 2)
```

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -1049,12 +1049,54 @@ trait SourceCompanionOps:
     */
   def future[T](from: Future[T])(using StageCapacity, ExecutionContext): Source[T] =
     val c = StageCapacity.newChannel[T]
-    from.onComplete {
-      case Success(value)                  => c.send(value); c.done()
-      case Failure(ex: ExecutionException) => c.error(ex.getCause)
-      case Failure(ex)                     => c.error(ex)
+    receiveAndSendFromFuture(from, c)
+    c
+
+  /** Creates a source that emits elements from future source when `from` completes or fails otherwise. The `from` completion is performed
+    * on the provided [[scala.concurrent.ExecutionContext]] whereas elements are emitted through Ox. Note that when `from` fails with
+    * [[scala.concurrent.ExecutionException]] then its cause is returned as source failure.
+    *
+    * @param from
+    *   A [[scala.concurrent.Future]] that returns source upon completion.
+    * @return
+    *   A source that will emit values upon a `from` [[scala.concurrent.Future]] completion.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   import scala.concurrent.ExecutionContext.Implicits.global
+    *   import scala.concurrent.Future
+    *
+    *   supervised {
+    *     Source
+    *       .futureSource(Future.failed(new RuntimeException("future failed")))
+    *       .receive()                                                           // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
+    *     Source.futureSource(Future.successful(Source.fromValues(1, 2))).toList // List(1, 2)
+    *   }
+    *   }}}
+    */
+  def futureSource[T](from: Future[Source[T]])(using Ox, StageCapacity, ExecutionContext): Source[T] =
+    val c = StageCapacity.newChannel[T]
+    val transportChannel = StageCapacity.newChannel[Source[T]](using StageCapacity(1))
+
+    receiveAndSendFromFuture(from, transportChannel)
+
+    forkDaemon {
+      transportChannel.receive() match
+        case ChannelClosed.Done           => c.done()
+        case ChannelClosed.Error(r)       => c.error(r)
+        case source: Source[T] @unchecked => source.pipeTo(c)
     }
     c
+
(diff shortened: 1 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1029,6 +1029,48 @@ trait SourceCompanionOps:
...
+      case Success(source) =>
+        supervised {
+          repeatWhile {
+            source.receive() match
+              case ChannelClosed.Done     => c.done(); false
+              case ChannelClosed.Error(r) => c.error(r); false
+              case t: T @unchecked        => c.send(t); true
+          }
+        }
+      case Failure(ex: ExecutionContext) => c.error(ex.getCause)
```
adamw, line 1069: see the comment on Source.future :)

core/src/test/scala/ox/channels/SourceOpsFutureSourceTest.scala
```diff
@@ -0,0 +1,28 @@
...
+class SourceOpsFutureSourceTest extends AnyFlatSpec with Matchers {
+  import scala.concurrent.ExecutionContext.Implicits.global
+
+  behavior of "SourceOps.futureSource"
+
+  it should "fail to create source from the null future" in supervised {
+    the[IllegalArgumentException] thrownBy {
+      Source.futureSource(null)
+    } should have message "requirement failed: from cannot be null"
+  }
```
geminicaprograms, line 18: this is no longer needed

core/src/main/scala/ox/channels/SourceOps.scala
geminicaprograms: fixed

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -993,6 +995,82 @@ trait SourceCompanionOps:
...
+    *   supervised {
+    *     Source
+    *       .futureSource(Future.failed(new RuntimeException("future failed")))
+    *       .receive()                                                           // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
+    *     Source.futureSource(Future.successful(Source.fromValues(1, 2))).toList // List(1, 2)
+    *   }
+    *   }}}
+    */
+  def futureSource[T](from: Future[Source[T]])(using Ox, StageCapacity, ExecutionContext): Source[T] =
+    require(from != null, "from cannot be null")
```
geminicaprograms, line 1057: removed `null` check

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1028,6 +1028,47 @@ trait SourceCompanionOps:
...
+    *       .receive()                                                           // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
+    *     Source.futureSource(Future.successful(Source.fromValues(1, 2))).toList // List(1, 2)
+    *   }
+    *   }}}
+    */
+  def futureSource[T](from: Future[Source[T]])(using Ox, StageCapacity, ExecutionContext): Source[T] =
+    val c = StageCapacity.newChannel[T]
+    from.onComplete {
+      case Success(source) =>
+        supervised {
```
rucek, line 1059: **question:**
1. Can we use `source.pipeTo(c)` instead?
2. Do we need `supervised` here? We're not starting any forks that would need supervision.

geminicaprograms: 1. yes, great idea 💪

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1028,6 +1028,39 @@ trait SourceCompanionOps:
...
+    *       .futureSource(Future.failed(new RuntimeException("future failed")))
+    *       .receive()                                                           // ChannelClosed.Error(Some(java.lang.RuntimeException: future failed))
+    *     Source.futureSource(Future.successful(Source.fromValues(1, 2))).toList // List(1, 2)
+    *   }
+    *   }}}
+    */
+  def futureSource[T](from: Future[Source[T]])(using StageCapacity, ExecutionContext): Source[T] =
+    val c = StageCapacity.newChannel[T]
+    from.onComplete {
+      case Success(source)                 => supervised { source.pipeTo(c) }
```
adamw, line 1058: we block in `onComplete`, which probably shouldn't happen - the pipe should only complete once all elements have been transmitted

geminicaprograms: that's correct, 🤔 the other option would be to somehow pass the received (from `onComplete -> Success`) `Source` to another `fork` and process it there so that emission doesn't happen in the `ExecutionContext`. Let me think and propose sth there...

adamw: yes, I think adding a source-transferring channel (with a buffer size of 1) would do the trick (plus a fork awaiting for the channel)

geminicaprograms: WDYT?

adamw:
Thanks :)

PR #47, opened 2023-11-02
feat: implement `orElse` operator

### LINKED ISSUE #37: Add `orElse` operator

geminicaprograms:
If a source has no elements then elements from an `alternative` source are emitted to the returned channel. If this source is failed then failure is passed to the returned channel.

Examples:
```scala
  Source.fromValues(1).orElse(Source.fromValues(2, 3)).toList // List(1)
  Source.empty.orElse(Source.fromValues(2, 3)).toList         // List(2, 3)
```

Closes #37.

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -779,6 +779,34 @@ trait SourceOps[+T] { this: Source[T] =>
         }
         buffer.result()
       }
+
+  /** If this source has no elements then elements from an `alternative` source are emitted to the returned channel. If this source is
+    * failed then failure is passed to the returned channel.
+    *
+    * @param alternative
+    *   An alternative source of elements used when this source is empty.
+    * @return
+    *   A source that emits either elements from this source or from `alternative` (when this source is empty).
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     Source.fromValues(1).orElse(Source.fromValues(2, 3)).toList // List(1)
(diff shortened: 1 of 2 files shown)


adamw (approved):

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -779,6 +779,34 @@ trait SourceOps[+T] { this: Source[T] =>
...
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     Source.fromValues(1).orElse(Source.fromValues(2, 3)).toList // List(1)
+    *     Source.empty.orElse(Source.fromValues(2, 3)).toList         // List(2, 3)
+    *   }
+    *   }}}
+    */
+  def orElse[U >: T](alternative: Source[U])(using Ox, StageCapacity): Source[U] =
+    val c = StageCapacity.newChannel[U]
```
rucek, line 802: **question:** Do we need a new channel here? Conceptually this seems equivalent to 
```scala
if isDone then alternative else this
```
which satisfies all the test cases, but I (or the test cases) might be missing something.

adamw: the channel might only become done after the first receive - e.g. if the producer sleeps, and the completes the channel, we'll only find out after we've completed the first `receive()`

geminicaprograms: yeap. here is the case that fails with the proposed solution but work with the existing code
```scala
Source.fromValues[Int]().orElse(Source.fromValues(2, 3)).toList shouldBe List(2, 3)
```
I have added it to the test cases so that nobody is tempted to over-optimise 😃

PR #52, opened 2023-11-22
mapPar for collections

amorfis:
Allows code like in README. Also adds extension to `Iterable` with `.mapParWith`. 

There is ugly `asInstanceOf` in the code, any ideas how to do it better are welcome

### DIFF
--- a/README.md
+++ b/README.md
@@ -56,7 +56,8 @@ val result: List[Int] = mapPar(input)(4)(_ + 1)
 // (2, 3, 4, 5, 6, 7, 8, 9, 10)
 ```
 
-If any transformation fails, others are interrupted and `mapPar` throws exception. Parallelism
+If any transformation fails, others are interrupted and `mapPar` rethrows exception that was
+thrown by the transformation. Parallelism
 limits how many concurrent forks are going to process the collection.
 
 ## Race two computations
--- a/README.md
+++ b/README.md
@@ -45,6 +45,20 @@ val result: (Int, String) = par(computation1)(computation2)
 
 If one of the computations fails, the other is interrupted, and `par` waits until both branches complete.
 
+## Parallelize collection transformation
+
+```scala
+import ox.mapPar
+
+val input: List[Int] = List(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+
+val result: List[Int] = mapPar(input)(4)(_ + 1)
+// (2, 3, 4, 5, 6, 7, 8, 9, 10)
+```
+
+If any transformation fails, others are interrupted and `mapPar` throws exception. Parallelism
+limits how many concurrent forks are going to process the collection.
+
 ## Race two computations
 
 ```scala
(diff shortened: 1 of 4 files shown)


core/src/main/scala/ox/mapPar.scala
```diff
@@ -0,0 +1,13 @@
+package ox
+
+import scala.collection.IterableFactory
+
+def mapPar[I, O, C[E] <: Iterable[E]](parallelism: Int)(iterable: => C[I])(transform: I => O): C[O] =
```
adamw, line 5: it would be good to add some scaladoc; I'm also not sure if the implementation is optimal - if the elements in the first batch take really long to process, and in the second complete quickly, we'll end up with a single thread doing the mapping. In theory, the second thread could then take over the work

rucek:
Woult it make sense to re-use `Source.mapPar` for this, like:
```scala
val res = Source.fromIterable(iterable).mapPar(parallelism)(f)
```
and then transforming `res` to the expected `Iterable` using the provided `IterableFactory`

README.md
```diff
@@ -45,6 +45,20 @@ val result: (Int, String) = par(computation1)(computation2)
...
+```scala
+import ox.mapPar
+
+val input: List[Int] = List(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+
+val result: List[Int] = mapPar(input)(4)(_ + 1)
+// (2, 3, 4, 5, 6, 7, 8, 9, 10)
+```
+
+If any transformation fails, others are interrupted and `mapPar` throws exception. Parallelism
```
ritschwumm, line 59: "throws exception" sounds wrong - "throws an exception" maybe? apart from that - some specific kind of exception? as a developer, i'd like to know.

amorfis:
@rucek I tried it, but it's ~3 times slower then directly using semaphores and forks

core/src/main/scala/ox/mapPar.scala
amorfis: I agree with the suboptimal implementation. It is changed now. Scaladoc also added

adamw:
@amorfis @rucek yes making streams behave better is what I'm working on now :) but with Source.fromIterator there's always an extra fork

core/src/main/scala/ox/syntax.scala
```diff
@@ -24,3 +26,6 @@ object syntax:
     def useInScope: T = ox.useCloseableInScope(f)
     def useScoped[U](p: T => U): U = ox.useScoped(f)(p)
     def useSupervised[U](p: T => U): U = ox.useSupervised(f)(p)
+
+  extension [I, C[E] <: Iterable[E]](f: => C[I])
+    def mapParWith[O](parallelism: Int)(transform: I => O) = ox.mapPar(parallelism)(f)(transform)
```
adamw, line 31: any reason to use `mapParWith` vs `mapPar`?

adamw (approved):

core/src/main/scala/ox/syntax.scala
adamw: ah, that's because `f1().parWith(f2())` reads better - you're running the receiver & argument in parallel. Here, each argument has a different role, so I think we can simply have `mapPar`

PR #53, opened 2023-11-22
Parse test reports in the GH Actions workflow

amorfis:
Adds "Test results" under the jobs in the workflow:

Screenshot 2023-11-22 at 18 09 14

Makes it easier to find failed tests and presents nice overview of all the tests run

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,6 +23,13 @@ jobs:
       run: sbt -v compile
     - name: Test
       run: sbt -v test
+    - name: Test Report
+      uses: dorny/test-reporter@v1
+      if: success() || failure()
+      with:
+        name: 'Tests results'
+        path: '**/test-reports/TEST*.xml'
+        reporter: java-junit
 
   publish:
     name: Publish release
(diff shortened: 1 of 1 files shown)


adamw:
Nice! How come we didn't have this before? ;)

.github/workflows/ci.yml
```diff
@@ -23,6 +23,13 @@ jobs:
       run: sbt -v compile
     - name: Test
       run: sbt -v test
+    - name: Test Report
+      uses: dorny/test-reporter@v1
+      if: success() || failure()
```
adamw, line 28: isn't this a no-op? that is, can't we simply remove this?

amorfis: No, it means to run this step regardless if the previous one failed or succeeded. Without it test reports wouldn'd be parsed if any test failed.

adamw: Ah ok :) maybe add a comment, I think people might find it confusing (e.g. me ;) )

adamw:
Thanks!

PR #54, opened 2023-11-23
collectPar, filterPar, foreachPar for collections added

amorfis:
The tests became very repeatable. I'm wondering if it would be a good move to abstract them somehow.

Parallel processing is abstracted now in `commonPar` private[ox] method. Better name suggestions are welcome


README.md
```diff
@@ -45,21 +45,66 @@ val result: (Int, String) = par(computation1)(computation2)
...
 limits how many concurrent forks are going to process the collection.
 
+### foreachPar
+
+```scala
+import ox.syntax.foreachPar
+
+val input: List[Int] = List(1, 2, 3, 4, 5, 6, 7, 8, 9, 10)
+
+input.mapPar(4)(i => println())
```
adamw, line 72: foreach? :)

core/src/main/scala/ox/collectPar.scala
```diff
@@ -0,0 +1,32 @@
+package ox
+
+import scala.collection.IterableFactory
+
+/** Runs partial function in parallel on each element of `iterable` for which the partiel function is defined.
```
adamw, line 5: typo

core/src/test/scala/ox/util/MaxCounter.scala
```diff
@@ -0,0 +1,20 @@
...
+import java.util.concurrent.atomic.AtomicInteger
+
+class MaxCounter() {
+  val counter = new AtomicInteger(0)
+  var max = 0
+
+  def increment() = {
+    counter.updateAndGet { c =>
+      val inc = c + 1
+      max = if (inc > max) inc else max
```
adamw, line 12: this looks unsafe - using a mutable variable from multiple threads

amorfis: Indeed, I had doubts too, but it's only in test and well... it works. 

Moreover I can't think of scenario where it could do any harm. I.e. even if one thread updates the `max` var and then "rolls the transaction back" (as it's in optimistic locking) and the `max` remains updated - that means the other thread updates it to the same value anyway.

adamw: a thread might first read the the value, then another thread might bump max a couple of times, and then the original thread might write the incorrect (smaller) value back

one day you might see a weird test fail ;)

amorfis: How can it write a smaller value? In case of simultaneous access both threads will see the same value of `c` won't them? Even if one makes several failed attempts to update, it will always be "the old value" won't it?

adamw:
Thank you! :)

PR #99, opened 2024-03-12
Add Source.mapConcat operator

kciesielski:
Closes https://github.com/softwaremill/ox/issues/35

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -538,6 +538,52 @@ trait SourceOps[+T] { outer: Source[T] =>
     }
     c
 
+  /** Applies the given mapping function `f`, to each element received from this source, transforming it into an Iterable of results, then
+    * sends the results one by one to the returned channel. Can be used to unfold incoming sequences of elements into single elemnts.
+    *
+    * @param f
+    *   A function that transforms the element from this source into a pair of the next state into an [[scala.collection.IterableOnce]] of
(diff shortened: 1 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -538,6 +538,52 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+    *   }}}
+    */
+  def mapConcat[U](f: T => IterableOnce[U])(using Ox, StageCapacity): Source[U] =
+    val c = StageCapacity.newChannel[U]
+    fork {
+      repeatWhile {
+        receiveSafe() match
+          case ChannelClosed.Done =>
+            try c.doneSafe()
+            catch case t: Throwable => c.errorSafe(t)
```
adamw, line 570: doneSafe can't throw exceptions

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -538,6 +538,52 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+            false
+          case ChannelClosed.Error(r) =>
+            c.errorSafe(r)
+            false
+          case t: T @unchecked =>
+            try
+              val results: IterableOnce[U] = f(t)
+              results.iterator.map(c.sendSafe).forall(_.isValue)
```
adamw, line 578: I think we can just use .send here, the channel can only be closed by this method (since we are returning a Source), so it can't really be closed externally

adamw (approved):

Issue #56, opened 2023-12-04
Add side-effecting callback to retry attempts


rucek:
When the retry mechanism performs another attempt to run the operation, it might be useful to execute a side effect - like logging or updating metrics.

Some things to consider:
- how much context should such a callback have (e.g. which attempt is this)
- should the callback be executed before or after the attempt? In the latter case it could include the result, which - particularly in case of error - might be useful for logging/metrics. Or perhaps both such callbacks should be available, so that we can hook into the retry lifecycle even more?

PR #106, opened 2024-03-19
Add side effect callbacks during retry

DybekK:
This PR addresses the proposed functionality in https://github.com/softwaremill/ox/issues/56 and introduces enhancements to the retry mechanism by adding a `onRetry` function callback to the `RetryPolicy` class. 

`onRetry` callback is executed after each retry attempt. It accepts two parameters: the attempt number and the result of the attempt. The result is represented as an Either type, where Left indicates an error and Right indicates a successful result. By default, this is an empty function.


kciesielski (commented):
I don't think we need the `beforeEachAttempt` part, can't see any good use case for it. Users can simply enrich their retried effect with `effectBefore >> myEffect` and retry such composite effect. Adding this to the retry API could the flow a bit confusing. The `cats-retry` library offers a `onError: (E, RetryDetails) => M[Unit]` callback, and I think we should be fine with a similar thing. Wdyt @rucek @DybekK?

core/src/test/scala/ox/retry/RetryLifecycleTest.scala
```diff
@@ -0,0 +1,95 @@
...
+    // then
+    result shouldBe successfulResult
+    counter shouldBe 1
+
+    beforeEachAttemptInvoked shouldBe true
+    beforeEachAttemptInvocationCount shouldBe 1
+
+    afterEachAttemptInvoked shouldBe true
+    afterEachAttemptInvocationCount shouldBe 1
+    returnedResult shouldBe Right(successfulResult)
```
rucek, line 51: Aren't those assertions redundant? If you verify that the callback was invoked once, I guess one of them would be sufficient. Same for the "before" callback.

core/src/test/scala/ox/retry/RetryLifecycleTest.scala
```diff
@@ -0,0 +1,95 @@
...
+    // then
+    result shouldBe failedResult
+    counter shouldBe 4
+
+    beforeEachAttemptInvoked shouldBe true
+    beforeEachAttemptInvocationCount shouldBe 4
+
+    afterEachAttemptInvoked shouldBe true
+    afterEachAttemptInvocationCount shouldBe 4
+    returnedResult shouldBe Left(failedResult)
```
rucek, line 94: Also here the assertions seem redundant - wouldn't it be sufficient to verify the invocation count here? Same for the "before" callback.

core/src/main/scala/ox/retry/retry.scala
```diff
@@ -41,28 +52,36 @@ def retryEither[E, T](operation: => Either[E, T])(policy: RetryPolicy[E, T]): Ei
...
-def retry[E, F[_], T](em: ErrorMode[E, F])(operation: => F[T])(policy: RetryPolicy[E, T]): F[T] =
+def retry[E, F[_], T](em: ErrorMode[E, F])(operation: => F[T])(policy: RetryPolicy[E, T], lifecycle: RetryLifecycle[E, T]): F[T] =
   @tailrec
   def loop(attempt: Int, remainingAttempts: Option[Int], lastDelay: Option[FiniteDuration]): F[T] =
     def sleepIfNeeded =
       val delay = policy.schedule.nextDelay(attempt + 1, lastDelay).toMillis
       if (delay > 0) Thread.sleep(delay)
       delay
 
+    lifecycle.beforeEachAttempt(attempt + 1)
```
rucek, line 70: I'm wondering if we could just number the attempts starting from 1 instead of 0 - WDYT?

rucek (commented):
Some minor coments, but overall LGTM.

rucek:
> Users can simply enrich their retried effect with `effectBefore >> myEffect` and retry such composite effect

@kciesielski the difefrence would be that `effectBefore` won't be aware of the attempt number, but `beforeEachAttempt` will. Not sure about an actual use case for that though.

kciesielski:
Yes, I'm aware of that difference, but I can't see the usefulness of this attempt number in such a case. I think we should choose a simpler API and sacrifice it.

core/src/main/scala/ox/retry/retry.scala
DybekK: I agree. In my opinion, it is more intuitive to read the code if we get rid of `attempt + 1` in some places.

DybekK:
> I don't think we need the `beforeEachAttempt` part, can't see any good use case for it. Users can simply enrich their retried effect with `effectBefore >> myEffect` and retry such composite effect. Adding this to the retry API could the flow a bit confusing. The `cats-retry` library offers a `onError: (E, RetryDetails) => M[Unit]` callback, and I think we should be fine with a similar thing. Wdyt @rucek @DybekK?

I completely removed `RetryLifecycle` class and replaced it with just `onRetry` function callback.

core/src/main/scala/ox/retry/retry.scala
```diff
@@ -10,27 +10,49 @@ import scala.util.Try
...
   * @param policy
   *   The retry policy - see [[RetryPolicy]].
   * @return
   *   The result of the function if it eventually succeeds.
   * @throws anything
   *   The exception thrown by the last attempt if the policy decides to stop.
   */
-def retry[T](operation: => T)(policy: RetryPolicy[Throwable, T]): T =
-  retryEither(Try(operation).toEither)(policy).fold(throw _, identity)
+def retry[T](operation: => T, onRetry: OnRetry[Throwable, T] = emptyOnRetry)(policy: RetryPolicy[Throwable, T]): T =
```
adamw, line 22: maybe `onRetry` should be then part of the `RetryPolicy`? Its docs says that `A policy that defines how to retry a failed operation.` - I think this easily includes an `onRetry` callback?

rucek: I agree, I was thinking about that as well. Until we had two callbacks, it made sense to me to have them as a separate parameter, but as we have a single callback now, I'd make it a part of the `RetryPolicy` as well.

adamw:
Thanks!

Issue #58, opened 2023-12-04
Composable retry policies


rucek:
The current retry mechanism allows to provide a single policy for retrying the operation. Once this policy is "exhausted", i.e. when it's done retrying without success, it gives up.

We'd like to be able to compose multiple policies, so that more than one can be tried in the attempt to run the operation. For example, we might want to make a couple of immediate retries (with no delay) and, if still unsuccessful, retry indefinitely with exponential backoff. 

The above could be written with a hypothetical `fallbackTo` combinator like:
```scala
val policy: RetryPolicy = RetryPolicy.immediate(3).fallbackTo(RetryPolicy.backoffForever(100.millis))
```

zkerriga:
Hi, @rucek! I made solution to compose Schedules so we can construct policies like:
```scala
RetryPolicy(Schedule.Immediate(1).fallbackTo(Schedule.Delay(2, 2.seconds)))

RetryPolicy(Schedule.Delay(2, 1.second).fallbackTo(Schedule.Backoff.forever(2.seconds)))
```
but I am not sure about the design of it. Please, have a look at my request

PR #129, opened 2024-04-27
add ability to combine Schedules for retries

zkerriga:
#58 

I tried to make proposed syntax: 
```scala
val policy: RetryPolicy = RetryPolicy.immediate(3).fallbackTo(RetryPolicy.backoffForever(100.millis))
```
but found `fallbackTo` method on `RetryPolicy` a bit misleading, because with current design of retry classes we cannot easily compose different `ResultPolicy` and `onRetry` functions. Instead, I decided to make composable just `Schedule` classes.

I added `Combination` and `CombinationForever` subclasses by analogy so we can combine `Finite` schedules with both `Finite` and `Infinite` fallbacks, but I have a feeling that they add an unreasonable complication :thinking: - feel free to reject it 

P.S. it's my first opensource request

### DIFF
--- a/core/src/main/scala/ox/resilience/Schedule.scala
+++ b/core/src/main/scala/ox/resilience/Schedule.scala
@@ -10,6 +10,8 @@ object Schedule:
 
   private[resilience] sealed trait Finite extends Schedule:
     def maxRetries: Int
+    def fallbackTo(fallback: Finite): Finite = Combination(this, fallback)
+    def fallbackTo(fallback: Infinite): Infinite = Combination.forever(this, fallback)
 
   private[resilience] sealed trait Infinite extends Schedule
 
@@ -137,7 +139,3 @@ object Schedule:
     def forever(base: Finite, fallback: Infinite): Infinite = CombinationForever(base, fallback)
 
   case class CombinationForever private[resilience](base: Finite, fallback: Infinite) extends Combined, Infinite
-
-  extension (schedule: Finite)
-    def fallbackTo(fallback: Finite): Finite = Combination(schedule, fallback)
-    def fallbackTo(fallback: Infinite): Infinite = Combination.forever(schedule, fallback)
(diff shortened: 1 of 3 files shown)


core/src/main/scala/ox/resilience/Schedule.scala
```diff
@@ -117,3 +117,27 @@ object Schedule:
...
+
+  object Combination:
+    /** A schedule that retries indefinitely, using [[base]] first [[base.maxRetries]] number of times, and then always using [[fallback]].
+      */
+    def forever(base: Finite, fallback: Infinite): Infinite = CombinationForever(base, fallback)
+
+  case class CombinationForever private[resilience](base: Finite, fallback: Infinite) extends Combined, Infinite
+
+  extension (schedule: Finite)
+    def fallbackTo(fallback: Finite): Finite = Combination(schedule, fallback)
```
kciesielski, line 142: **suggestion:** If this is the same file, maybe you could just put these into the `Finite` trait? I don't think we'd mind such a coupling.

rucek:
Thanks for the contribution @zkerriga! This looks good too me, my only concerns are about the naming, which is indeed hard 😅. I think `Combined` and `Combination` are too general, since what we're doing is not a generic combination of `Schedule`s, but a very specific one - a fallback to another `Schedule`.

I was thinking about renaming `Combined` to `WithFallback`, and `Combination`/`CombinationForever` to `FallingBack`/`FallingBackForever`, respectively. WDYT?

kciesielski:
`FallingBackForever` sounds so romantic <3

adamw:
Since we are bikeshedding on names (the PR looks good, @zkerriga , let's get it merged & released :) ), how about simply `andThen` instead of `fallbackTo`? Wouldn't it be more intuitive, i.e., "retry 3 times, and then retry 5 times with a pause". The class representing the schedule could then be named `Sequence` (as there's a "first" and "second" schedule, that is, order matters)?

zkerriga:
Thank you for reviews! @adamw, it's an interesting idea about `Sequence` and `andThen` method, but I decided to use `Fallback` naming from @rucek because it seems clearer to me and I don't think that making more than 1 fallback will be a popular thing.

rucek:
I agree that although `andThen` and `Sequence` would make sense, it's the "fallback" naming that communicates the purpose more clearly.

rucek (approved):

PR #133, opened 2024-05-08
Update retries.md

scalway:
retry basic syntax is different than shown in first snippet


adamw:
Indeed, that one snippet is not compile. Thank you! :)

PR #134, opened 2024-05-09
Add Source <> InputStream operations and chunks

kciesielski:


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -679,4 +679,6 @@ trait SourceOps[+T] { outer: Source[T] =>
         case t: T @unchecked        => c.sendOrClosed(t); pipeTo(c)
     }
     c
+
+  def through[U](f: Ox ?=> Source[T] => Source[U])(using Ox): Source[U] = f(this)
```
kciesielski, line 683: I think we should define a `Pipe[T, U]` alias for such a function.

core/src/main/scala/ox/Chunk.scala
```diff
@@ -0,0 +1,55 @@
+package ox
+
+import scala.reflect.{ClassTag, classTag}
+
+abstract sealed class Chunk[+A]:
```
adamw, line 5: 1. conceptually, what is a chunk? what are its defining characteristics, that make it different e.g. from `List`, `Vector` or `Iterable`? And once we know our true motivations here, it would be good to put it in a comment :)

2. can a chunk be part of scala collections e.g. by implementing `IterableOnce` or sth like that?

core/src/main/scala/ox/channels/SourceCompanionOps.scala
```diff
@@ -383,3 +387,44 @@ trait SourceCompanionOps:
     val c = Channel.rendezvous[T]
     c.errorOrClosed(t)
     c
+
+  def fromInputStream(is: InputStream, chunkSize: Int = 1024)(using Ox): Source[Chunk[Byte]] =
```
adamw, line 391: to keep the class from growing, maybe it would make sense to put this e.g. in a `SourceCompanionIOOps`? just for code organization

core/src/main/scala/ox/channels/SourceCompanionOps.scala
```diff
@@ -383,3 +387,44 @@ trait SourceCompanionOps:
...
+          if r == -1 then
+            chunks.done()
+            false
+          else
+            val chunk = if r == chunkSize then Chunk.fromArray(a) else Chunk.fromArray(a.take(r))
+            chunks.send(chunk)
+            true
+        }
+      catch
+        case t: Throwable =>
```
adamw, line 407: NonFatal! :)

adamw: we need a linter for that ;)

adamw: ah no sorry 🤦 stream operators catch IE on purpose ... as these are propagated downstream. I discover this not the first time, I just need to document it somewhere visible ... ;)

core/src/main/scala/ox/channels/SourceCompanionOps.scala
```diff
@@ -383,3 +387,44 @@ trait SourceCompanionOps:
...
+          chunks.errorOrClosed(t)
+      finally
+        try is.close()
+        catch case _: Throwable => ()
+    }
+    chunks
+
+extension (source: Source[Chunk[Byte]]) {
```
adamw, line 415: alternative impl: put this in `SourceOps` (or `SourceIOOps`), as a regular method, and require a witness in signature that the `T =:= Chunk[Byte]`.

Pros: `.asInputStream` is always visible in auto-complete, without the `ox.streams.*` import. Cons: probably some, but I don't know now ;)

core/src/main/scala/ox/channels/text.scala
```diff
@@ -0,0 +1,24 @@
+package ox.channels
+
+import ox.*
+import scala.annotation.tailrec
+
+object text {
+  val lines: Ox ?=> Source[Chunk[Byte]] => Source[String] = source =>
```
adamw, line 7: wouldn't this rather be

```
trait SourceOps[T}:
  def lines(using Ox, T =:= Chunk[Byte]): Source[String] = ...
```

?

adamw: Or `SourceOps.io.text` & `SourceOps.io.asInputStream` etc.?

core/src/main/scala/ox/channels/SourceOps.scala
adamw: what's the use-case? :)

core/src/main/scala/ox/channels/SourceCompanionOps.scala
kciesielski: This, and also I copied that from your code :) https://github.com/softwaremill/sttp/issues/2158

core/src/main/scala/ox/channels/SourceOps.scala
kciesielski: There can be more pipes in the future, not to mention reuse in Tapir's websockets ;) A named abstraction would make this easier to reason about (`through(pipe)`), this is inspired by analogous approach in fs2.

core/src/main/scala/ox/channels/text.scala
adamw: that is, `SourceOps.text.lines` etc

core/src/main/scala/ox/channels/SourceOps.scala
adamw: yeah, but what's the use case of capturing a `Ox ?=> ...` function, and passing it to a `(using Ox)` method? Otherwise we have `.pipe` ;)

core/src/main/scala/ox/Chunk.scala
kciesielski: I will definitely add a detailed Scaladoc. The main idea is to allow streaming lightweight array-like structures which are cheap to operate on. With arrays, concatenation, drop, take, and similar operations are expensive, as they copy elements. Streaming libraries like fs2, zio, and akka-streams have come up with their representations with underlying optimizations.
From user's perspective, the interface is very similar to `Array` and yes, we can make `Chunk` extend IndexedSeq[A]`.
For a start, underlying implementation uses Arrays, which can be further optimized later without changing types exposed to the user.

core/src/main/scala/ox/channels/SourceCompanionOps.scala
kciesielski: Nice, thanks for the hint! This trick makes extension methods a bit less cool, it would be nice to have them resolvable automatically if they are, for example, in the same package as the extended type.

core/src/main/scala/ox/channels/SourceOps.scala
kciesielski: Hmm of course! Not sure what I was imagining by putting them both together :)

core/src/main/scala/ox/Chunk.scala
adamw: Yeah, still I think it's good to have at least some sketch of a motivation on why a given data type is introduced, to be certain that we actually need a new one :)

core/src/main/scala/ox/channels/text.scala
kciesielski: @adamw you mean something like this? https://github.com/softwaremill/ox/pull/134/files#diff-289bd5bbd44b1c2e8ce41f9f27847cd6170983360562f2d188206b8fff869118R13

adamw: yeah, what do you think? In `Source.fromValues(chunk).text.lines.toList`, the `.text` looks a bit like a property. So maybe simply `Source.fromValues(chunk).lines.toList` would be better?

core/src/main/scala/ox/channels/SourceCompanionOps.scala
kciesielski: One more thing that comes to my mind is how would relying on Ops traits to such an extent affect binary compatibility in the future? Adding new combinators requires adding new traits or extending existing Ops traits, which may break bincompat? In comparison, exposing a `through(pipe)` interface and defining pipes externally as functions wouldn't cause such problems, as well as using extension methods.
I'm asking just out of curiosity, fully realizing that bincompat is not a concern at this point.

adamw: I don't think adding a new trait would cause any problems bin-compat wise? Same as adding methods.

As for `.through(pipe)`, it might be a useful abstraction, but I still need to see a specific use-case ;)

core/src/main/scala/ox/channels/text.scala
kciesielski: Yeah, this 'text' looked a bit excessive to me here as well. I'll leavy only `.lines`.

core/src/main/scala/ox/Chunk.scala
```diff
@@ -0,0 +1,63 @@
+package ox
+
+import scala.reflect.{ClassTag, classTag}
+
+/** An immutable finite indexed sequence of elements. Backed by Array, meant to be used as a standard abstraction in Sources where passing
+  * around Array is not desirable. Standard operations like concatenation, drop, splitAt, etc. used in Source processing can be used just
+  * like with arrays, but their internal implementation can be optimized to avoid unnecessary expensive data copying.
```
adamw, line 7: you cover the implementation (backed by an array), but you don't write what's special about `Chunk` - as I understand, the main reason to use chunks is because of performance. Still, why not just use arrays then? It would be useful to include that here

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
```diff
@@ -0,0 +1,38 @@
...
+            val chunk = if r == chunkSize then Chunk.fromArray(a) else Chunk.fromArray(a.take(r))
+            chunks.send(chunk)
+            true
+        }
+      catch
+        case t: Throwable =>
+          chunks.errorOrClosed(t)
+      finally
+        try is.close()
+        catch case _: Throwable => ()
```
adamw, line 36: we can also then try to report the error?

core/src/main/scala/ox/Chunk.scala
kciesielski: Arrays are pretty bad performance-wise when you want to do operations like split, drop, take, and concatenation. These are operations that will typically take place in stream processing. I can rephrase the scaladoc to emphasize that it's about performance (and immutability as a bonus, but that would be covered by Scala 3 IArray if one insisted, so I don't mention that explicitly).

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
kciesielski: You mean something like printStackTrace?

adamw: no, `chunks.errorOrClosed(t)` :)

core/src/main/scala/ox/Chunk.scala
adamw: yes, I'd put exactly that in the scaladoc :)

core/src/main/scala/ox/Chunk.scala
```diff
@@ -0,0 +1,65 @@
+package ox
+
+import scala.reflect.{ClassTag, classTag}
+
+/** An immutable finite indexed sequence of elements, backed by Array. Raw rrays are expensive performance-wise when you want to do
+  * operations like concatenation, splitAt, drop, etc. Such operations are often useful when doing Source processing. `Chunk` offers a
+  * wrapper focused primarly on performance optimizations for such costly operations, while maintaining familiar API of an `IndexedSeq`.
+  * Underneath, a `Chunk` should leverage lazy data structures to avoid unnecessary data copying, which is a typical drawback for most Array
```
adamw, line 8: it does? it looks like a wrapper on an array, are there any optimizations in palce?

adamw (approved):
Apart from chunk - looks good :) I mean, chunk is ok as well, but I'm not sure if it's complete

core/src/main/scala/ox/Chunk.scala
kciesielski: There are no optimizations yet, that's why I wrote that it "should" use these optimizations. I intended this PR to expose the Chunk API but start with simple non-optimized array operations. It can then later evolved as needed.

adamw: Ok, sounds good, let's just explicitly write that: Chunk for now is a think wrapper on arrays, in the future we want to optimize its operations

adamw:
Thanks :) Can you add follow-up tasks to (a) optimize the chunks, (b) add more input/output methods, I think reading/writing to/from `File`/`Path` would be a good start, anything else that you think is essential? Maybe also check with fs2/akka streams for inspiration

kciesielski:
Chunk task: https://github.com/softwaremill/ox/issues/136
As for other operations I'd consider essential: 
* It may be useful to have a `_.chunk` method for `Source[Byte]`, especially if we want to represent byte streams like this in Tapir. Conversely - `_.unchunk`.
* I/O: File read/write, maybe also `Sink.fromOutputStream`?
* Text: we could consider adding methods to decode/encode between `Source[Chunk[Byte]]`  and `Source[String]` with given charset (UTF-8 for a start). 
* Reactive streams integration

adamw:
> I/O: File read/write, maybe also Sink.fromOutputStream?

Yes, or just source.toOutputStream(OutputStream), same as you might have source.toFile(File) (or Path). We could also write  a `Source[String]` to a file, just as a `Source[Chunk[Byte]]`.

> Text: we could consider adding methods to decode/encode between Source[Chunk[Byte]] and Source[String] with given charset (UTF-8 for a start).

I guess `.lines` is such a method? 

But let's open the tasks, we can discuss there further.

kciesielski:
Ok, tasks created:
https://github.com/softwaremill/ox/issues/138
https://github.com/softwaremill/ox/issues/137

I think encode/decode will be a different thing than .lines, not assuming line breaks but working on all kinds of text streams continually.

adamw:
Thanks :) Ok, we'll need to document this properly then.

Issue #137, opened 2024-05-14
Add Source.fromFile and source.toFile


kciesielski:
1. The `source.toFile` should work for `Source[Chunk[Byte]]` and `Source[String]`.
2. Add Source.toOutputStream as well

PR #139, opened 2024-05-14
File I/O

kciesielski:
Closes https://github.com/softwaremill/ox/issues/137


core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
```diff
@@ -40,3 +47,58 @@ trait SourceCompanionIOOps:
...
+      *
+      * @param path
+      *   path the file to read from.
+      * @param chunkSize
+      *   maximum number of bytes to read from the file before emitting a new chunk.
+      * @return
+      *   a `Source` of chunks of bytes.
+      * @throws IOException
+      *   If an I/O error occurs when opening the file.
+      * @throws SecurityException
```
adamw, line 61: isn't the whole security manager deprecated in java?

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
```diff
@@ -40,3 +47,58 @@ trait SourceCompanionIOOps:
...
+    if Files.isDirectory(path) then throw new IOException(s"Path $path is a directory")
+    val chunks = StageCapacity.newChannel[Chunk[Byte]]
+    val jFileChannel = useInScope {
+      try {
+        FileChannel.open(path, StandardOpenOption.READ)
+      } catch
+        case _: UnsupportedOperationException =>
+          // Some file systems don't support file channels
+          Files.newByteChannel(path, StandardOpenOption.READ)
+    }(_.close())
```
adamw, line 74: close is idempotent?

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
```diff
@@ -40,3 +47,58 @@ trait SourceCompanionIOOps:
...
+    fork {
+      repeatWhile {
+        val buf = ByteBuffer.allocate(chunkSize)
+        try {
+          val readBytes = jFileChannel.read(buf)
+          if readBytes < 0 then
+            try
+              jFileChannel.close()
```
adamw, line 82: move this to `finally`?

core/src/main/scala/ox/channels/SourceIOOps.scala
```diff
@@ -22,3 +29,37 @@ trait SourceIOOps[+T]:
...
+    if Files.isDirectory(path) then throw new IOException(s"Path $path is a directory")
+    val jFileChannel = 
+      try {
+        FileChannel.open(path, StandardOpenOption.WRITE)
+      } catch
+        case _: UnsupportedOperationException =>
+          // Some file systems don't support file channels
+          Files.newByteChannel(path, StandardOpenOption.WRITE)
+    
+    def closeJFileChannel(cause: Option[Throwable]): Unit =
```
adamw, line 43: this can probalby be a private inline util method, also usable if fromFile?

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
kciesielski: I don't know TBH, just put it here because javadoc for FileChannel [mentions this](https://github.com/openjdk/jdk/blob/4ba74475d44831c1fe49359458163cd1567e9619/src/java.base/share/classes/java/nio/channels/FileChannel.java#L348) as a possible `@throws`.

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
kciesielski: Yeah, I though abot that, and that's why I have put `close` additionally in other places. This one serves just as a "last resort", and it's supposed to be idempotent according to the javadoc.

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
kciesielski: I'm not sure if I get it. The only thing that's under `try` here is closing the channel when `readBytes < 0`.

core/src/main/scala/ox/channels/SourceIOOps.scala
kciesielski: There's a bit different handling, in `fromFile` we pass the exception to `chunks.errorOrClosed(ex)`, here we rethrow it. Still, it may be worth inlining.

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
adamw: yes, but the close handler remains in the scope. If the scope is long-lived, the registered close handlers might accumulate

core/src/main/scala/ox/channels/SourceIOOps.scala
adamw: ah ok :)

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
adamw: the close can be in `finally`, so that the `catch case NonFatal(closeException) =>` is not duplicated?

kciesielski: Ok, I think I get it -> I have to move `try` out, to make it around `repeatWhile` and then it indeed would be possible, thanks.

doc/index.md
```diff
@@ -69,8 +69,9 @@ In addition to this documentation, ScalaDocs can be browsed at [https://javadoc.
 
 .. toctree::
    :maxdepth: 2
-   :caption: Kafka integration
+   :caption: Integrations
```
kciesielski, line 72: This would look like:

image

Or maybe we want to keep Kafka at top level?

adamw: Kafka can be a subsection, that's not a problem, though I'm not sure if I would put file i/o under "integrations" - it's built-in, not an additional module. Since it concerns channels only maybe putting it under "channels" would make more sense?

doc/flake.nix
```diff
@@ -0,0 +1,32 @@
+{
+  description = "Python shell flake";
+
+  inputs = {
+    nixpkgs.url = "github:nixos/nixpkgs/nixos-unstable";
+  };
+
+  outputs = { self, nixpkgs }: {
+
+    devShell.x86_64-linux = let
```
kciesielski, line 10: I just noticed this is OS-dependent, fixing.

adamw:
Thanks :)

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
```diff
@@ -40,3 +47,50 @@ trait SourceCompanionIOOps:
...
+      * @return
+      *   a `Source` of chunks of bytes.
+      * @throws IOException
+      *   If an I/O error occurs when opening the file.
+      * @throws SecurityException
+      *   If SecurityManager error occurs when opening the file.
+      */
+  def fromFile(path: Path, chunkSize: Int = 1024)(using Ox): Source[Chunk[Byte]] =
+    if Files.isDirectory(path) then throw new IOException(s"Path $path is a directory")
+    val chunks = StageCapacity.newChannel[Chunk[Byte]]
```
adamw, line 66: btw @kciesielski here we could have an implicit `StageCapacity` as well

core/src/main/scala/ox/channels/SourceCompanionIOOps.scala
```diff
@@ -40,3 +47,50 @@ trait SourceCompanionIOOps:
...
+    fork {
+      try {
+        repeatWhile {
+          val buf = ByteBuffer.allocate(chunkSize)
+          val readBytes = jFileChannel.read(buf)
+          if readBytes < 0 then
+            chunks.done()
+            false
+          else if readBytes == 0 then
+            chunks.send(Chunk.empty)
```
adamw, line 83: and re-reading this, is there any point in sending empty chunks? ;)

kciesielski: Right, no point :)

kciesielski: @adamw PR with fixes to this https://github.com/softwaremill/ox/pull/143

PR #140, opened 2024-05-15
IO capability & compiler plugin

adamw:


doc/io.md
```diff
@@ -0,0 +1,123 @@
...
+
+In other words, the presence of a `using IO` parameter indicates that the method might:
+
+* have side effects: write to a file, send a network request, read from a database, etc.
+* take a non-trivial amount of time to complete due to blocking, data transfer, etc.
+* throw an exception (unless the exceptions are handled, and e.g. transformed into 
+  [application errors](basics/error-handling.md))
+
+Quite importantly, the **absence** of `using IO` specifies that the method has **no** I/O side effects (however, it 
+might still block the thread, e.g. when using a channel, or have other side effects, such as throwing exceptions).
```
kciesielski, line 24: > or have other side effects, such as throwing exceptions
or interact with randomness and system time?

adamw: yes, I don't think that would count as IO :)

doc/io.md
```diff
@@ -0,0 +1,123 @@
...
+Compiler assists in checking this property, but only to a certain degree - it's possible to cheat!
+
+The `IO` capability can be introduced using `IO.unsafe`. Ideally, this method should only be used at the edges of your 
+application (e.g. in the `main` method), or when integrating with third-party libraries. Otherwise, the capability
+should be passed as an implicit parameter. Such an ideal scenario might not possible, but there's still value in `IO`
+tracking: by looking up the usages of `IO.unsafe` it's possible to quickly find the "roots" where the `IO` capability 
+is introduced. For example:
```
kciesielski, line 31: Are there cases when it won't be in some kind of main method which puts together all the modules?

adamw: I was thinking e.g. of callbacks when integrating with reactive-streams or such?

doc/io.md
kciesielski: I was thinking if it's worth mentioning here as 'other side effects'.

doc/io.md
```diff
@@ -0,0 +1,123 @@
...
+
+Take care not to capture the capability e.g. using constructors (unless you are sure such usage is safe), as this might 
+circumvent the tracking of I/O operations. Similarly, the capability might be captured by lambdas, which might later be 
+used when the IO capability is not in scope. In future Scala and Ox releases, these problems should be detected at 
+compile-time using the upcoming capture checker.
+
+## The requireIO compiler plugin
+
+Ox provides a compiler plugin, which verifies at compile-time that the `IO` capability is present when invoking any
+methods from the JDK or Java libraries that specify to throw an `java.io.IOException`.
```
kciesielski, line 64: Did you consider including `java.sql.SQLException`? JDBC methods throw it instead of `IOException` for communication errors, and it's part of the JDK. On the other hand, users probably interact with libraries that, even if wrapping JDBC, don't necessarily throw raw `SQLException`. I saw some signs of this happening here and there in ScalikeJDBC, but that may not be enough to justify the idea :)

doc/io.md
adamw: ah yes, good idea :)

doc/io.md
adamw: Yeah good idea, the list is already extensible, we could easily provide a good default.

PR #148, opened 2024-06-04
Proposal: prevent either block nesting to avoid type/try-catch mismatches

lbialy:
Either blocks, based on boundaries, provide implicit evidence for safe jumps. One problematic case is when `.ok()` combinator is called on a fork (or inside of exception capturing scope that can swallow a Break), another is when nested boundaries provide evidences for mismatched cases:

```scala
val effectMatchingOuterEither: Either[Throwable, Unit] = Left(Exception("oh no"))
val effectMatchingInnerEither: Either[String, Int] = Right(3)

val outer: Either[Throwable, Unit] = either {
  val inner: Either[String, Int] = either {
    effectMatchingOuterEither.ok()
    effectMatchingInnerEither.ok()
  }

  ()
}
 ```
 
Notice that first `Either` value evaluated with `.ok()` matches outer `either` block while being evaluated in the inner block. This might very well be considered a feature - types allow us to decide where to jump - but in my short but stressful experience with this kind of combinators - when this is used with type inference, results can be surprising. In my case the inner block was the body of Tapir handler while outer either block was in main function. The result was, unsurprisingly, a thread killed with `Break`. I would get a compilation error if nested either blocks were not allowed because without the outer block I'd be informed that it's impossible to use `.ok()` as `either` block in tapir handler is inferred to be `Either[Unit, String]`. I know this is a rather heavy-handed change and there are possible use cases for nested either blocks (tagged jumps for example) but this is confusing enough imho to warrant such a change. I can easily imagine such a change to occur during refactoring of code where inner either block stops matching one of the eithers with `.ok()` in it's scope and outer either providing evidence where it's no longer handled correctly.

### DIFF
--- a/core/src/main/scala/ox/either.scala
+++ b/core/src/main/scala/ox/either.scala
@@ -1,11 +1,15 @@
 package ox
 
+import scala.annotation.implicitNotFound
 import scala.compiletime.{error, summonFrom}
-import scala.util.boundary
+import scala.util.{NotGiven, boundary}
 import scala.util.boundary.{Label, break}
 import scala.util.control.NonFatal
 
 object either:
+
+  private type NotNested = NotGiven[Label[Either[Nothing, Nothing]]]
+
   /** Within an [[either]] block, allows unwrapping [[Either]] and [[Option]] values using [[ok()]]. The result is the right-value of an
     * `Either`, or the defined-value of the `Option`. In case a failure is encountered (a left-value of an `Either`, or a `None`), the
     * computation is short-circuited and the failure becomes the result. Failures can also be reported using [[fail()]].
@@ -32,7 +36,11 @@ object either:
     *       v1.ok() ++ v2.ok()
     *   }}}
     */
-  inline def apply[E, A](inline body: Label[Either[E, A]] ?=> A): Either[E, A] = boundary(Right(body))
+  inline def apply[E, A](inline body: Label[Either[E, A]] ?=> A)(using
+      @implicitNotFound(
+        "Nesting of either blocks is not allowed as it's bug prone with type inference. Consider extracting the nested either block to a separate function."
+      ) nn: NotNested
+  ): Either[E, A] = boundary(Right(body))
 
   extension [E, A](inline t: Either[E, A])
     /** Unwrap the value of the `Either`, short-circuiting the computation to the enclosing [[either]], in case this is a left-value. */
--- a/core/src/test/scala/ox/EitherTest.scala
+++ b/core/src/test/scala/ox/EitherTest.scala
@@ -5,6 +5,8 @@ import org.scalatest.flatspec.AnyFlatSpec
 import org.scalatest.matchers.should.Matchers
 import ox.either.{fail, ok}
 
+import scala.util.boundary.Label
+
 class EitherTest extends AnyFlatSpec with Matchers:
   val ok1: Either[Int, String] = Right("x")
   val ok2: Either[Int, String] = Right("y")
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by Ichoran** [comment]
In my experience, once one is aware that this can happen, it's actually a nice feature.  Normally, yes, you jump to the closest block.  But you can get out farther if you need to.  It's unusual--at least for me--to be so confused that I have both the type wrong and where to jump to wrong, but they accidentally align and work despite not being desired behavior.

The problem comes when you try to jump thread boundaries.  So, from my experience at least, I think this avoidance of nesting is not desirable.

**Comment by lbialy** [comment]
In simplistic examples, sure. The problem I encountered was in a block in which tapir endpoints were also declared, either block was opened on top for `Either[Throwable, Unit]` and inferred type of endpoint handler was `Either[Unit, String]`. I had a function that had `Either[Throwable, String]` that I `.ok()`ed in handler and it caught the top-level either handler instead of the one inside of the handler so two things happened simultaneously: either blocks overlapped and there was a thread boundary that won't be covered by #147. Now disallowing nested either blocks won't fix this case completely and neither will #147 because you can always do:
```scala
either:
  val t = Thread(() => e.ok())
  t.start()
  ```
but I still feel it would prevent bugs coming from refactoring that are possible due to the automatic application of implicits.

File: core/src/main/scala/ox/either.scala [hunk 1]
```diff
@@ -32,7 +36,11 @@ object either:
     *       v1.ok() ++ v2.ok()
     *   }}}
     */
-  inline def apply[E, A](inline body: Label[Either[E, A]] ?=> A): Either[E, A] = boundary(Right(body))
+  inline def apply[E, A](inline body: Label[Either[E, A]] ?=> A)(using
+      @implicitNotFound(
+        "Nesting of either blocks is not allowed as it's bug prone with type inference. Consider extracting the nested either block to a separate function."
```
adamw, line 41: Maybe reformulate a bit? Sth like:

```
Nesting of either blocks is not allowed as it's error prone, due to type inference. Consider extracting the nested either block to a separate function.
```

adamw:
Looks good - let's try and see when we get complaints :) Do you think we should mention this in the docs?

lbialy:
I guess we should with an example on how to refactor.

adamw:
@lbialy Would you be able to add the docs as well? :)

lbialy:
Yeah. Btw: I was thinking whether there should be a `either.nested:` block that would allow for nested either blocks (maybe either.unsafe?).

adamw:
> Yeah. Btw: I was thinking whether there should be a either.nested: block that would allow for nested either blocks (maybe either.unsafe?).

Let's be lazy here and wiat for the first user complaint that they would like such behavior ;)

lbialy:
Ok, I did add docs and changed the error message to your suggestion.

doc/basics/error-handling.md
```diff
@@ -120,3 +120,55 @@ import ox.catching
...
+
+Proper way to solve this is to extract the inner `either:` block to a separate function:
+
+```scala mdoc:compile-only
+import ox.either, either.*
+
+def returnsEither: Either[Exception, Int] = ???
+
+def inner(): Either[String, Int] = either:
+  val i = returnsEither.ok() // this can only jump to either on the opening of this function
```
adamw, line 168: this does not compile (as indicated by the CI failure) - you can't `.ok()` here as the boundary is `String`-error-typed, and `returnsEither` is `Exception`-error-typed.

adamw:
Ok, one merged, couple more to go :) Thanks :)

PR #151, opened 2024-06-05
Add includeFailed to takeWhile

kciesielski:
A useful feature.

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -294,8 +294,8 @@ trait SourceOps[+T] { outer: Source[T] =>
           case ChannelClosed.Done =>
             c.doneOrClosed().discard
             false
-          case ChannelClosed.Error(r) =>
-            c.errorOrClosed(r).discard
+          case ChannelClosed.Error(reason) =>
+            c.errorOrClosed(reason).discard
             false
           case t: T @unchecked =>
             try
@@ -305,8 +305,8 @@ trait SourceOps[+T] { outer: Source[T] =>
                 c.doneOrClosed().discard
                 false
             catch
-              case NonFatal(e) =>
-                c.errorOrClosed(e).discard
+              case t: Throwable =>
+                c.errorOrClosed(t).discard
                 false
       }
     }


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -277,11 +281,36 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+  def takeWhile(f: T => Boolean, includeFailed: Boolean = false)(using Ox, StageCapacity): Source[T] =
+    val c = StageCapacity.newChannel[T]
+    fork {
+      repeatWhile {
+        receiveOrClosed() match
+          case ChannelClosed.Done =>
+            c.doneOrClosed().discard
+            false
+          case ChannelClosed.Error(r) =>
+            c.errorOrClosed(r).discard
```
kciesielski, line 298: **question:** Should we ignore if erroring the channel wasn't successful? (here, and in the catch below as well). This is unexpected, but in case of bugs we'd lose  the original error. I saw such approach in other ox functions.

adamw: If the channel is already in an error, the original error will remain. If that's the case, `.errorOrClose` will return the original one, and do nothing else

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -277,11 +281,36 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+            false
+          case t: T @unchecked =>
+            try
+              if f(t) then c.sendOrClosed(t).isValue
+              else
+                if includeFailed then c.sendOrClosed(t).discard
+                c.doneOrClosed().discard
+                false
+            catch
+              case NonFatal(e) =>
```
adamw, line 308: shouldn't we catch `Throwable` here as in other operators?

adamw: (not to close the scope, just propagate the error)

core/src/main/scala/ox/channels/SourceOps.scala
kciesielski: > we'd lose the original error.
sorry, bad wording, I didn't mean the original error but the "next error". I guess the only alternative is to combine errors, throw and tear down the entire scope, which sounds like a bad idea in this case.

BUT, apart from this case, in case of the `channel.error(e)` implementation, throwing original error and ignoring `e` could be improved to throwing original error with `e` added to its suppressed exceptions?

adamw: But then we could run into concurrency issues. I don't think `.addSuppressed` is thread-safe?

PR #157, opened 2024-06-20
first draft of OxApp with extension companion traits

lbialy:


lbialy:
@adamw CI crash is unrelated to this code 😮

adamw:
@lbialy yes, already fixed the test: https://github.com/softwaremill/ox/commit/7a4a73b52ba945ca490a6451a0ac1b334f234440

core/src/test/scala/ox/OxAppTest.scala
```diff
@@ -0,0 +1,254 @@
...
+  "OxApp" should "work in happy case" in {
+    var ec = Int.MinValue
+
+    object Main1 extends OxApp:
+      override def exit(code: Int): Unit =
+        ec = code
+
+      def run(args: Vector[String])(using Ox): ExitCode = Success
+
+    Main1.main(Array.empty)
```
adamw, line 22: we could also assert that the result is 0 here

core/src/main/scala/ox/OxApp.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  private[ox] def exit(code: Int): Unit = System.exit(code)
+
+  private[ox] def mountShutdownHook(thread: Thread): Unit =
+    try Runtime.getRuntime.addShutdownHook(thread)
+    catch case _: IllegalStateException => ()
+
+  private[ox] def printStackTrace(t: Throwable): Unit = t.printStackTrace()
+
+  private[OxApp] final def handleRun(args: Vector[String])(using Ox): ExitCode =
```
adamw, line 37: is `Ox` needed here?

core/src/main/scala/ox/OxApp.scala
```diff
@@ -0,0 +1,65 @@
...
+      interruptThread.setName("ox-interrupt-hook")
+
+      mountShutdownHook(interruptThread)
+
+      cancellableMainFork.joinEither() match
+        case Left(iex: InterruptedException) => exit(0)
+        case Left(fatalErr)                  => throw fatalErr
+        case Right(exitCode)                 => exit(exitCode.code)
+
+  private[ox] def exit(code: Int): Unit = System.exit(code)
```
adamw, line 29: maybe add a comment that these are `private[ox]` vs `private` because of testing

core/src/main/scala/ox/OxApp.scala
```diff
@@ -0,0 +1,65 @@
...
+
+  private[ox] def mountShutdownHook(thread: Thread): Unit =
+    try Runtime.getRuntime.addShutdownHook(thread)
+    catch case _: IllegalStateException => ()
```
adamw, line 33: shouldn't we re-throw the exception if mounding the hook fails?

core/src/main/scala/ox/OxApp.scala
```diff
@@ -0,0 +1,65 @@
...
+        cancellableMainFork.cancel()
+        ()
+      })
+
+      interruptThread.setName("ox-interrupt-hook")
+
+      mountShutdownHook(interruptThread)
+
+      cancellableMainFork.joinEither() match
+        case Left(iex: InterruptedException) => exit(0)
```
adamw, line 25: is this success, if the app was interrupted?

core/src/main/scala/ox/OxApp.scala
adamw: ah we need it for `run`, ok

core/src/main/scala/ox/OxApp.scala
```diff
@@ -0,0 +1,65 @@
...
+
+object OxApp:
+  trait Simple extends OxApp:
+    override final def run(args: Vector[String])(using Ox): ExitCode =
+      run
+      ExitCode.Success
+
+    def run(using Ox): Unit
+
+  trait WithErrors[E] extends OxApp:
```
adamw, line 54: can we use an arbitrary error mode `WithErrorMode`, and specialise to eithers with `WithEither`?

adamw:
Looks great! :) Some docs would be good :)

core/src/main/scala/ox/OxApp.scala
lbialy: sure

core/src/main/scala/ox/OxApp.scala
lbialy: highly debatable, some sources treat interruption as failure and return 1 or 130 (128+SIGINT(2)), some return 0 if application was able to interrupt itself and shut down cleanly (as there were no issues during shutdown), it's up to us to decide. JVM itself returns 130 on unhandled sigint.

core/src/main/scala/ox/OxApp.scala
lbialy: sure

core/src/main/scala/ox/OxApp.scala
lbialy: for reference: CE IOApp only shuts down it's thread pools when it gets ISE here

core/src/test/scala/ox/OxAppTest.scala
lbialy: sure

core/src/main/scala/ox/OxApp.scala
lbialy: is something like:
```

  trait WithErrorMode[E, F[_]](em: ErrorMode[E, F]) extends OxApp:
    override def run(args: Vector[String])(using ox: Ox): ExitCode =
      val result = run(args.toList)
      if em.isError(result) then ExitCode.Failure(1)
      else ExitCode.Success

    def run(args: List[String])(using Ox): F[ExitCode]

  abstract class WithEitherErrors[E] extends WithErrorMode(EitherMode[E]()):

    type EitherScope[Err] = Label[Either[Err, ExitCode]]

    override final def run(args: List[String])(using ox: Ox): Either[E, ExitCode] =
      either[E, ExitCode](label ?=> run(args.toVector)(using ox, label))

    def run(args: Vector[String])(using Ox, EitherScope[E]): ExitCode
 ```
 what you had in mind?

Comment (lbialy) [reply core/src/main/scala/ox/OxApp.scala:54]: mind the change of Vector to List in args because we have conflicting signatures so there's that, we could do a disambiguation by always-provided implicit but that's even more... arcane

also without handleErrors there's no translation of error into exit code

Comment (lbialy) [reply core/src/main/scala/ox/OxApp.scala:54]: it could be also:
```scala
  trait WithErrorMode[E, F[_]](em: ErrorMode[E, F]) extends OxApp:
    override def run(args: Vector[String])(using ox: Ox): ExitCode =
      val result = run(args.toList)
      if em.isError(result) then
        try handleError(em.getError(result))
        catch
          case NonFatal(e) =>
            e.printStackTrace()
            ExitCode.Failure(1)
      else ExitCode.Success

    def handleError(e: E): ExitCode

    def run(args: List[String])(using Ox): F[ExitCode]
 ```

Comment (lbialy) [reply core/src/main/scala/ox/OxApp.scala:54]: also notice the abstract class, traits are not allowed to provide constructor args for traits that are their supers

Comment (lbialy) [reply core/src/main/scala/ox/OxApp.scala:54]: and `F[ExitCode]` doesn't make much sense in case where we have `Label[Either[E, Nothing]]` in scope as it means that errors are to be delivered with `.ok()` / `break()` so unwrapped `ExitCode` as retval

File: core/src/test/scala/ox/OxAppTest.scala (hunk 1 above)
Comment (lbialy) [reply core/src/test/scala/ox/OxAppTest.scala:22]: but main is `Unit`, we check ec to be equal 0 below

File: core/src/main/scala/ox/OxApp.scala (hunk 5 above)
Comment (adamw) [reply core/src/main/scala/ox/OxApp.scala:25]: ok, so here we would treat this as "handled sigint"?

but that's great info, can you put the above in a comment in the code? for future readers :)

File: core/src/main/scala/ox/OxApp.scala (hunk 6 above)
Comment (adamw) [reply core/src/main/scala/ox/OxApp.scala:54]: hm the Vector/List is also hacky. Maybe we simply need a different name. We've got `fork` & `forkError` for the fork-with-error-mode variant,  maybe here we could keep a similar convention? (`run` & `runError`). Or come up with better names for both :)

another nitpick: I'd avoid using "scope" in names, as we already have concurrency scopes etc. Maybe rename `EitherScope` -> `EitherError`?

Comment (adamw) [reply core/src/main/scala/ox/OxApp.scala:54]: also, does `: F[ErrorCode]` and `ErrorCode` have the same erasure? Isn't the first erased to `Object`, making the signatures different?

File: core/src/main/scala/ox/OxApp.scala [hunk 7]
```diff
@@ -0,0 +1,65 @@
...
+
+import scala.util.boundary.*
+import scala.util.control.NonFatal
+
+enum ExitCode(val code: Int):
+  case Success extends ExitCode(0)
+  case Failure(exitCode: Int = 1) extends ExitCode(exitCode)
+
+trait OxApp:
+  def main(args: Array[String]): Unit =
```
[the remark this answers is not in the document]
lbialy: true

core/src/main/scala/ox/OxApp.scala
lbialy: ok, I'll even make this configurable

core/src/main/scala/ox/OxApp.scala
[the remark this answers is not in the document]
lbialy: yeah, it does

core/src/main/scala/ox/OxApp.scala
adamw: Ah right, signature is only name+parameters, not the return type

lbialy:
The signature that requires `List[String]` is only present in WithErrorMode trait and is expected to be used in the child trait so yeah, can be renamed to runWithErrors.

adamw:
@lbialy yeah this should work fine, also will be consistent with `supervisedError` and `forkError` that we already have

lbialy:
@adamw this is now ready :)

hearnadam (approved):

core/src/main/scala/ox/OxApp.scala
```diff
@@ -0,0 +1,143 @@
...
+
+}
+
+object OxApp {
+
+  case class AppSettings(
+      /** This value is returned to the operating system as the exit code when the app receives SIGINT and shuts itself down gracefully.
+        * Default value is `ExitCode.Success` (0). JVM itself returns code `130` when it receives `SIGINT`.
+        */
+      gracefulShutdownExitCode: ExitCode = ExitCode.Success
```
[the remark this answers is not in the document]
adamw: Yes, since we have the default instance let's move this there

PR #158, opened 2024-06-21
Add Source.grouped* operators

### LINKED ISSUE #34: Add `grouped` (with variants) operator
Check [grouped](https://doc.akka.io/docs/akka/current/stream/operators/Source-or-Flow/grouped.html) for inspiration

micossow:
closes #34

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -716,4 +716,29 @@ trait SourceOps[+T] { outer: Source[T] =>
         case t: T @unchecked        => c.sendOrClosed(t); pipeTo(c)
     }
     c
+
+  /** TODO: documentation
+    */
+  def grouped(n: Int)(using Ox, StageCapacity): Source[Seq[T]] =
+    val c2 = StageCapacity.newChannel[Seq[T]]
+    val buffer = scala.collection.mutable.Buffer.empty[T]
+    fork {
+      repeatWhile {
+        receiveOrClosed() match
+          case ChannelClosed.Done =>
+            if buffer.nonEmpty then
+              c2.sendOrClosed(buffer.toSeq); ()
+            c2.doneOrClosed(); false
+          case ChannelClosed.Error(r) =>
+            c2.errorOrClosed(r); false
+          case t: T @unchecked =>
+            buffer += t
+            if buffer.size == n then
+              val isValue = c2.sendOrClosed(buffer.toSeq).isValue
+              buffer.clear()
+              isValue
+            else true
+      }
+    }
+    c2
 }
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -11,7 +11,7 @@ import scala.collection.IterableOnce
 import scala.concurrent.duration.*
 import scala.util.{Failure, Success, Try}
 
-case class GroupingTimeout(marker: Long)
+case class GroupingTimeout()
 
 trait SourceOps[+T] { outer: Source[T] =>
   // view ops (lazy)
@@ -751,7 +751,6 @@ trait SourceOps[+T] { outer: Source[T] =>
   def groupedWithin(n: Int, duration: FiniteDuration)(using Ox, StageCapacity): Source[Seq[T]] =
     val c2 = StageCapacity.newChannel[Seq[T]]
     val timerChannel = StageCapacity.newChannel[GroupingTimeout]
-    val currentTimeoutMarker = new AtomicLong(0)
     fork {
       var buffer = Vector.empty[T]
       repeatWhile {
@@ -764,15 +763,13 @@ trait SourceOps[+T] { outer: Source[T] =>
             false
           case ChannelClosed.Error(r) =>
             c2.errorOrClosed(r); false
-          case timerChannel.Received(GroupingTimeout(timeoutMarker)) =>
-            if timeoutMarker == currentTimeoutMarker.get() then
-              // no new element has been received since the fork was created - sending grouped elements
+          case timerChannel.Received(GroupingTimeout()) =>
+            // timeout - sending all elements from the buffer
+            if buffer.nonEmpty then
               val isValue = c2.sendOrClosed(buffer).isValue
               buffer = Vector.empty
               isValue
-            else
-              // a new element has been received since the fork was created - ignoring this timeout
-              true
+            else true
           case Received(t) =>
             buffer = buffer :+ t
             if buffer.size == n then
@@ -784,7 +781,7 @@ trait SourceOps[+T] { outer: Source[T] =>
               // buffer is not full - enqueuing marker to the timeout channel
               fork {
                 sleep(duration)
-                timerChannel.sendOrClosed(GroupingTimeout(currentTimeoutMarker.incrementAndGet()))
+                timerChannel.sendOrClosed(GroupingTimeout())
               }
               true
       }
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -9,8 +9,7 @@ import java.util.concurrent.{CountDownLatch, Semaphore}
 import scala.collection.IterableOnce
 import scala.concurrent.duration.*
 
-private[channels] trait Timeout
-private[channels] case object GroupingTimeout extends Timeout
+private[channels] case object GroupingTimeout
 
 trait SourceOps[+T] { outer: Source[T] =>
   // view ops (lazy)
@@ -852,7 +851,7 @@ trait SourceOps[+T] { outer: Source[T] =>
     */
   def groupedWeightedWithin(minWeight: Long, duration: FiniteDuration)(costFn: T => Long)(using Ox, StageCapacity): Source[Seq[T]] =
     val c2 = StageCapacity.newChannel[Seq[T]]
-    val timerChannel = StageCapacity.newChannel[Timeout]
+    val timerChannel = StageCapacity.newChannel[GroupingTimeout.type]
     fork {
       var buffer = Vector.empty[T]
       var accumulatedCost = 0L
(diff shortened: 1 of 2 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,178 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+    val c2 = StageCapacity.newChannel[Seq[T]]
+    val timerChannel = StageCapacity.newChannel[GroupingTimeout]
+    fork {
+      var buffer = Vector.empty[T]
+      var accumulatedCost = 0L
+      repeatWhile {
+        selectOrClosed(receiveClause, timerChannel.receiveClause) match
+          case ChannelClosed.Done =>
+            if buffer.nonEmpty then
+              c2.sendOrClosed(buffer)
```
kciesielski, line 862: **suggestion:** 
```scala
if buffer.nonEmpty then c2.sendOrClosed(buffer).discard
```

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,178 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+            accumulatedCost += costFn(t)
+            if accumulatedCost >= minWeight then
+              // buffer is full - sending grouped elements
+              val isValue = c2.sendOrClosed(buffer).isValue
+              buffer = Vector.empty
+              accumulatedCost = 0
+              isValue
+            else
+              // buffer is not full - enqueuing timeout to the separate channel
+              fork {
```
kciesielski, line 887: **issue:** You are starting the fork here if `accumulatedCost < minWeght`, so there can be multiple forked sleeps running when the buffer is filling up, right? I think we'd rather want just a single timeout ticking when the buffer is filling.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -9,7 +9,8 @@ import java.util.concurrent.{CountDownLatch, Semaphore}
 import scala.collection.IterableOnce
 import scala.concurrent.duration.*
 
-case class GroupingTimeout()
+private[channels] trait Timeout
```
kciesielski, line 12: **suggestion:** You don't need this base trait.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,179 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+        timerChannel.sendOrClosed(GroupingTimeout).discard
+      }
+      var timeoutFork: Option[CancellableFork[Unit]] = Some(forkTimeout())
+
+      def sendBufferAndForkNewTimeout(): Boolean =
+        val isValue = c2.sendOrClosed(buffer).isValue
+        buffer = Vector.empty
+        accumulatedCost = 0
+        timeoutFork.foreach(_.cancelNow())
+        timeoutFork = Some(forkTimeout())
```
kciesielski, line 870: **suggestion:** This can be conditional on `if isValue`

core/src/test/scala/ox/channels/SourceOpsGroupedTest.scala
```diff
@@ -0,0 +1,149 @@
...
+      c.send(1)
+      c.send(2)
+      c.send(3)
+      sleep(50.millis)
+      c.send(4)
+      sleep(200.millis) // to ensure the timeout is executed before the channel closes
+      c.done()
+    }
+    val elementsWithEmittedTimeOffset =
+      c.groupedWithin(3, 100.millis).map(s => (s, FiniteDuration(System.currentTimeMillis() - start, "ms"))).toList
```
kciesielski, line 56: **suggestion:** It's better to use `System.nanoTime` for measuring elapsed time, because it's guaranteed to be monotonic. https://medium.com/@batuhankadioglu/why-you-should-not-use-currenttimemillis-when-you-calculate-elapsed-time-d65ae36e764b

core/src/test/scala/ox/channels/SourceOpsGroupedTest.scala
```diff
@@ -0,0 +1,149 @@
...
+  }
+
+  it should "send the group only once when the channel is closed" in supervised {
+    val c = StageCapacity.newChannel[Int]
+    fork {
+      c.send(1)
+      c.send(2)
+      c.done()
+    }
+    c.groupedWithin(3, 100.millis).toList shouldBe List(List(1, 2))
```
kciesielski, line 120: **question:** This can emit due to `c.done()` or due to `100.millis`. Do I understand correctly that this test checks specifically that the result is emitted on `c.done()`? 
If yes, then how about something like
```scala
Try(timeout(30.seconds)(c.groupedWithin(3, 5.minutes).toList)) shouldBe Success(List(List(1, 2)))
```

core/src/main/scala/ox/channels/SourceOps.scala
micossow: suggestion applied

core/src/test/scala/ox/channels/SourceOpsGroupedTest.scala
micossow: suggestion applied

core/src/test/scala/ox/channels/SourceOpsGroupedTest.scala
micossow: suggestion applied

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,180 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+        timerChannel.sendOrClosed(GroupingTimeout).discard
+      }
+      var timeoutFork: Option[CancellableFork[Unit]] = Some(forkTimeout())
+
+      def sendBufferAndForkNewTimeout(): Boolean =
+        val isValue = c2.sendOrClosed(buffer).isValue
+        if isValue then
+          buffer = Vector.empty
+          accumulatedCost = 0
+          timeoutFork.foreach(_.cancelNow())
```
kciesielski, line 870: **suggestion:** We should cancel even if `isValue` is false, just start new fork if it's true.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,180 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+          case ChannelClosed.Error(r) =>
+            timeoutFork.foreach(_.cancelNow())
+            c2.errorOrClosed(r); false
+          case timerChannel.Received(GroupingTimeout) =>
+            timeoutFork = None // enter 'timed out state', may stay in this state if buffer is empty
+            if buffer.nonEmpty then sendBufferAndForkNewTimeout()
+            else true
+          case Received(t) =>
+            buffer = buffer :+ t
+            accumulatedCost += costFn(t)
```
kciesielski, line 890: **thought:** Since `costFn` is provided by the user, maybe we should handle it with `try/catch`? Then `c2.errorOrClosed` with catched `NonFatal` exception, cancel the fork and gently exit the while loop?

core/src/main/scala/ox/channels/SourceOps.scala
micossow: right, I made the fork conditional only - buffer should look empty when channel closes

core/src/main/scala/ox/channels/SourceOps.scala
micossow: I used the same try/catch pattern as in other operators (these catch Throwable)

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,196 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+    *       c.done()
+    *     }
+    *     c.groupedWithin(3, 100.millis).toList // List(Seq(1, 2), Seq(3, 4, 5), Seq(6))
+    *   }
+    *   }}}
+    */
+  def groupedWithin(n: Int, duration: FiniteDuration)(using Ox, StageCapacity): Source[Seq[T]] = groupedWeightedWithin(n, duration)(_ => 1)
+
+  /** Chunks up the elements into groups received within a time window or limited by the cumulative weight being greater or equal to the
+    * `minWeight`, whatever happens first. Timeout is counted since the last group has been emitted or no timeout is being counted. If this
```
kciesielski, line 828: **suggestion:** The timeout is reset after a group is emitted. If timeout expires and the buffer is empty, nothing is emitted. As soon as a new element is received, the source will emit it as a single-element group and reset the timer.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,196 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+    * source is failed then failure is passed to the returned channel.
```
kciesielski, line 829: **suggestion:** `If this source is failed...` => Upstream error or exception thrown by `costFn` will result in this `Source` failing with that error.

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,196 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+  def groupedWeighted(minWeight: Long)(costFn: T => Long)(using Ox, StageCapacity): Source[Seq[T]] =
+    val c2 = StageCapacity.newChannel[Seq[T]]
+    fork {
+      var buffer = Vector.empty[T]
+      var accumulatedCost = 0L
+      repeatWhile {
+        receiveOrClosed() match
+          case ChannelClosed.Done =>
+            if buffer.nonEmpty then c2.sendOrClosed(buffer).discard
+            c2.doneOrClosed();
```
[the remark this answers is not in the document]
micossow: fixed

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -716,4 +718,196 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+            false
+          case ChannelClosed.Error(r) =>
+            c2.errorOrClosed(r); false
```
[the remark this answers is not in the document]
micossow: fixed

core/src/main/scala/ox/channels/SourceOps.scala
micossow: corrected

core/src/main/scala/ox/channels/SourceOps.scala
micossow: corrected

kciesielski (approved):

PR #166, opened 2024-06-27
Source.alsoTo and alsoToTap operators

### LINKED ISSUE #39: Add `wireTap` operator

### LINKED ISSUE #40: Add `alsoTo` operator

micossow:
closes #39
closes #40

### DIFF
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -1035,11 +1035,14 @@ trait SourceOps[+T] { outer: Source[T] =>
     }
     c2
 
-  def wireTap[U >: T](other: Sink[U])(using Ox, StageCapacity): Source[U] =
+  case object Ack
+  def wireTap[U >: T](other: Sink[U], maxPending: Long = 3)(using Ox, StageCapacity): Source[U] =
+    val acks = StageCapacity.newChannel[Ack.type]
     val c2 = StageCapacity.newChannel[U]
     fork {
+      var pending: Int = 0
       repeatWhile {
-        receiveOrClosed() match
+        selectOrClosed(receiveClause, acks.receiveClause) match
           case ChannelClosed.Done =>
             c2.doneOrClosed()
             other.doneOrClosed()
@@ -1048,8 +1051,16 @@ trait SourceOps[+T] { outer: Source[T] =>
             c2.errorOrClosed(r)
             other.errorOrClosed(r)
             false
-          case t: T @unchecked =>
-            other.sendOrClosed(t).discard
+          case acks.Received(Ack) =>
+            pending -= 1
+            true
+          case Received(t) =>
+            if pending < maxPending then
+              pending += 1
+              fork {
+                other.sendOrClosed(t).discard
+                acks.sendOrClosed(Ack).discard
+              }.discard
             c2.sendOrClosed(t).isValue
       }
     }
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -10,6 +10,7 @@ import scala.collection.IterableOnce
 import scala.concurrent.duration.*
 
 private[channels] case object GroupingTimeout
+private[channels] case object Ack
 
 trait SourceOps[+T] { outer: Source[T] =>
   // view ops (lazy)
@@ -1009,6 +1010,27 @@ trait SourceOps[+T] { outer: Source[T] =>
     }
     c
 
+  /** Attaches the given Sink to this Source, meaning that elements that pass through will also be sent to the Sink. If the `other` Sink
+    * blocks, the returned channel also blocks.
+    *
+    * If this source is failed then failure is passed to the returned and channel and the `other` Sink.
+    *
+    * @param other
+    *   The Sink to which elements from this source will be sent.
+    * @return
+    *   A source that emits input elements.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     val c = Channel.withCapacity[Int](10)
+    *     Source.fromValues(1, 2, 3).alsoTo(c).toList // List(1, 2, 3)
+    *     c.toList // List(1, 2, 3)
+    *   }
+    *   }}}
+    */
   def alsoTo[U >: T](other: Sink[U])(using Ox, StageCapacity): Source[U] =
     val c2 = StageCapacity.newChannel[U]
     fork {
@@ -1016,7 +1038,7 @@ trait SourceOps[+T] { outer: Source[T] =>
         receiveOrClosed() match
           case ChannelClosed.Done =>
             c2.doneOrClosed()
-            other.doneOrClosed()
+            other.doneOrClosed() // TODO: should we close other channel?
             false
           case ChannelClosed.Error(r) =>
             c2.errorOrClosed(r)
@@ -1029,14 +1051,38 @@ trait SourceOps[+T] { outer: Source[T] =>
                 c2.doneOrClosed().discard
                 false
             else
-              other.doneOrClosed().discard
+              other.doneOrClosed().discard // TODO: should we close other channel?
               false
       }
     }
     c2
 
-  case object Ack
-  def wireTap[U >: T](other: Sink[U], maxPending: Long = 3)(using Ox, StageCapacity): Source[U] =
+  /** Attaches the given Sink to this Source, meaning that elements that pass through will also be sent to the Sink. The elements sent to
+    * `other` Sink are enqueued, which means that the returned channel does not block when the `other` Sink is blocked. If the queue is
+    * full, the elements are still sent to returned channel, but not to the `other` Sink, meaning that some elements may be dropped.
+    *
+    * If this source is failed then failure is passed to the returned and channel and the `other` Sink.
+    *
+    * @param other
+    *   The Sink to which elements from this source will be sent.
+    * @param queueSize
+    *   Specifies the size of the queue that buffers elements sent to the `other` Sink.
+    * @return
+    *   A source that emits input elements.
+    * @example
+    *   {{{
+    *   import ox.*
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     val c = Channel.withCapacity[Int](10)
+    *     Source.fromValues(1, 2, 3).wireTap(c).toList // List(1, 2, 3)
+    *     c.toList // List(1, 2, 3) but not guaranteed
+    *   }
+    *   }}}
+    */
+  def wireTap[U >: T](other: Sink[U], queueSize: Long = 3)(using Ox, StageCapacity): Source[U] =
+    require(queueSize > 0, "queueSize must be > 0")
     val acks = StageCapacity.newChannel[Ack.type]
     val c2 = StageCapacity.newChannel[U]
     fork {
@@ -1063,7 +1109,7 @@ trait SourceOps[+T] { outer: Source[T] =>
             pending -= 1
             true
           case Received(t) =>
-            if pending < maxPending then
+            if pending < queueSize then
               pending += 1
               fork {
                 other.sendOrClosed(t).discard
--- a/core/src/main/scala/ox/channels/SourceOps.scala
+++ b/core/src/main/scala/ox/channels/SourceOps.scala
@@ -1010,8 +1010,9 @@ trait SourceOps[+T] { outer: Source[T] =>
     }
     c
 
-  /** Attaches the given Sink to this Source, meaning that elements that pass through will also be sent to the Sink. If the `other` Sink
-    * blocks, the returned channel also blocks.
+  /** Attaches the given Sink to this Source, meaning that elements that pass through will also be sent to the Sink. If sending to the
+    * output channel or the `other` Sink blocks, no elements will be processed until both channels can receive elements again. The source
+    * elements are first sent to the output channel and then, only if the sent is successful, to the `other` Sink.
     *
     * If this source is failed then failure is passed to the returned and channel and the `other` Sink.
     *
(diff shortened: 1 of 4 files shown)


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1008,4 +1009,114 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+            c2.doneOrClosed()
+            // drain ack channel to ensure that all enqueued elements are sent
+            if pending != 0 then
+              repeatWhile {
+                acks.receiveOrClosed() match
+                  case ChannelClosed.Done     => false
+                  case ChannelClosed.Error(r) => false
+                  case Ack                    => pending -= 1; pending != 0
+              }
+              other.doneOrClosed().discard // TODO: should we close the other channel?
```
kciesielski, line 1102: Just as in `alsoTo` - yes, I'd see a strong dependency in this direction.
However, I'm not sure about having such a dependency in the other direction for `Done`. There is an interesting discussion about fs2's `observe`: https://github.com/typelevel/fs2/issues/1206. It's quite dated, but there are examples mentioned where the observing sink wants to finish early (like taking only a certain number of elements from `this`) without affecting the _main_ stream. Ultimately this logic hasn't been implemented. On the other hand, the documentation for akka's `wireTap` says that 
> elements that pass through will also be sent to the wire-tap Sink, without the latter affecting the mainline flow

so maybe it's worth testing how it behaves when `Sink` finishes early? My intuition is that it might be useful indeed to not close `this` if `Sink` closes.

micossow: I checked this behavior with the example below. `sideSink` receives `onUpstreamFinish` signal twice, but it's also materialized twice. I'm not sure if we can compare OX and Pekko 1:1

```scala
    class PrintingSink[T] extends GraphStage[SinkShape[T]] {
      val in: Inlet[T] = Inlet("PrintSink.in")
      override val shape: SinkShape[T] = SinkShape.of(in)

      override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
        println("Consumer starts")
        setHandler(
          in,
          new InHandler {
            override def onPush(): Unit = {
              val elem = grab(in)
              println(s"> $elem")
              pull(in)
            }

            override def onUpstreamFinish(): Unit = {
              println("Stream finished")
              completeStage()
            }

            override def onUpstreamFailure(ex: Throwable): Unit = {
              println(s"Stream failed with: ${ex.getMessage}")
              failStage(ex)
            }
          }
        )

        override def preStart(): Unit = {
          pull(in)
        }
      }
    }

    val source = AkkaSource(1 to 10)
    val source2 = AkkaSource(100 to 200)
    val sideSink = AkkaSink.fromGraph(PrintingSink[Int]())
    val mainSink = AkkaSink.foreach[Int](v => println(s">> $v"))

    val f = AkkaSource
      .combine(
        source.wireTap(sideSink),
        source2.wireTap(sideSink)
      )(i => Merge.apply(i))
      .runWith(mainSink)
```

micossow: With the modified example below, when we `completeStage()` after 2nd element, the `mainSink` keeps receiving elements until source finishes.
```scala
class PrintingSink[T] extends GraphStage[SinkShape[T]] {
      val in: Inlet[T] = Inlet("PrintSink.in")
      override val shape: SinkShape[T] = SinkShape.of(in)

      override def createLogic(inheritedAttributes: Attributes): GraphStageLogic = new GraphStageLogic(shape) {
        println("Consumer starts")
        setHandler(
          in,
          new InHandler {
            override def onPush(): Unit = {
              val elem = grab(in)
              println(s"> $elem")
              if elem == 2 then completeStage()
              else pull(in)
            }

            override def onUpstreamFinish(): Unit = {
              println("Stream finished")
              completeStage()
            }

            override def onUpstreamFailure(ex: Throwable): Unit = {
              println(s"Stream failed with: ${ex.getMessage}")
              failStage(ex)
            }
          }
        )

        override def preStart(): Unit = {
          pull(in)
        }
      }
    }

    val source = AkkaSource(1 to 10)
    val sideSink = AkkaSink.fromGraph(PrintingSink[Int]())
    val mainSink = AkkaSink.foreach[Int](v => println(s">> $v"))

    source.wireTap(sideSink).runWith(mainSink)
```
Output:
```
>> 1
> 1
>> 2
> 2
>> 3
>> 4
>> 5
>> 6
>> 7
>> 8
>> 9
>> 10
```

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1008,4 +1009,114 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+          case ChannelClosed.Done =>
+            c2.doneOrClosed()
+            other.doneOrClosed()
+            false
+          case ChannelClosed.Error(r) =>
+            c2.errorOrClosed(r)
+            other.errorOrClosed(r)
+            false
+          case t: T @unchecked =>
+            if c2.sendOrClosed(t).isValue then
```
adamw, line 1048: Shouldn't we propagate the closure reason to the other channel?

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1008,4 +1009,114 @@ trait SourceOps[+T] { outer: Source[T] =>
       }
     }
     c
+
+  /** Attaches the given Sink to this Source, meaning that elements that pass through will also be sent to the Sink. If the `other` Sink
+    * blocks, the returned channel also blocks.
```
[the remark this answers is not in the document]
adamw: It might also be worth noting, that we *first* send to `c2`, and only then to `other` - this ordering might be important

core/src/main/scala/ox/channels/SourceOps.scala
kciesielski: Good! I think it would be nice to have similar expectations in Ox - 
1. When `other` is `Done`, keep on sending to c2. I guess this is the current behavior, because the fork does `sendOrClosed(t).discard`. It might be useful to check what kind of Closed is returned and not send `Ack` if it was `Done`. This would fill up the queue and prevent further unneeded forks.
2. When `other` fails, fail `c2` as well. To be implemented in the mentioned `other.sendOrClosed(t)` result check .

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1008,4 +1009,114 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+    *   import ox.channels.Source
+    *
+    *   supervised {
+    *     val c = Channel.withCapacity[Int](10)
+    *     Source.fromValues(1, 2, 3).wireTap(c).toList // List(1, 2, 3)
+    *     c.toList // List(1, 2, 3) but not guaranteed
+    *   }
+    *   }}}
+    */
+  def wireTap[U >: T](other: Sink[U], queueSize: Long = 3)(using Ox, StageCapacity): Source[U] =
```
adamw, line 1084: so ... wireTap is like `alsoTo`, but with a buffer, and dropping elements if the buffer is full?

wouldn't a "proper" solution in such case be to just use an `other` sink with a larger capacity? Because it seems we are doing double-buffering here.

The option to drop elements if `other` is full might be useful of course, as an optional parameter to `alsoTo`?

kciesielski: Another difference is how the completion of `other` is handled. With `wireTap` we shouldn't care (as we're discussing here https://github.com/softwaremill/ox/pull/166#discussion_r1658456448). With `alsoTo` completion of `other` should complete `c2`.

adamw: ah good point :) still, these two look so similar, but have such different names ... ;)

core/src/main/scala/ox/channels/SourceOps.scala
micossow: Something like this?
```scala
          case t: T @unchecked =>
            c2.sendOrClosed(t) match
              case ChannelClosed.Done     => other.doneOrClosed().discard; false
              case ChannelClosed.Error(r) => other.errorOrClosed(r).discard; false
              case _ =>
                other.sendOrClosed(t) match
                  case ChannelClosed.Done     => c2.doneOrClosed().discard; false
                  case ChannelClosed.Error(r) => c2.errorOrClosed(r).discard; false
                  case _                      => true
```

core/src/main/scala/ox/channels/SourceOps.scala
micossow: > wouldn't a "proper" solution in such case be to just use an other sink with a larger capacity? 
I don't think so. Suppose we have an web application with attached external metric collecting system that goes down sometimes or is very slow, but we still want to process the requests as fast as possible. `wireTap` seems a good choice for this scenario, because it simply drops these metric updates instead of buffering them.

The proposed implementation introduces some kind of logical queue, but we could probably extend jox api with something like `c.sendOrIgnoreIfFull(t)`

adamw: `c.sendOrIgnoreIfFull == select(c.sendClause(t), Default(NotSent))`

core/src/main/scala/ox/channels/SourceOps.scala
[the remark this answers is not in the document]
micossow: rewording done

core/src/main/scala/ox/channels/SourceOps.scala
adamw: but we could also add this as a `trySend` method or sth similar with a name not suggesting exception-handling

core/src/main/scala/ox/channels/SourceOps.scala
adamw: yes, maybe we can capture this in a method, `propagateOrElse` or sth like this?

micossow: do you mean a method in the Sink trait?

adamw: no, just a helper method for streams impl

micossow: added `propagateOrElse`, does signature looks good to you?

core/src/main/scala/ox/channels/SourceOps.scala
micossow: this `select` is a game changer :)
one more thing regarding `wireTap`: should we propagate the error from `other` to `c2`?

micossow: cc @kciesielski

kciesielski: I assumed that we should, but Akka's wiretap documentation uses the expression 
> elements that pass through will also be sent to the wire-tap Sink, without the latter affecting the mainline flow
and now I'm not sure :) I think it's unsafe to let Sink fail silently. Could you test what happens in your test if you fail it on akka-streams after it processes a few elements?

micossow: > Could you test what happens in your test if you fail it on akka-streams after it processes a few elements?

after replacing `completeStage()` with `failStage(new RuntimeException("Boom!"))` in `wireTap` sink logic, it works the same way - elements pass through until source completes

adamw: Akka's model is a bit different because they have "cold" streams - they run the entire pipeline at once (when it's materialised). Here we've got "live" channels that we are operating on, which could be consumed by anybody anywhere. So if "tap" is supposed to be "best-effort", then if `other` is closed, I think we should just ignore it?

kciesielski: Hmm I understand this makes the attached secondary sink really not-affecting the mainline at all. Maybe that's what we need then, a true "silent listener". The `other` sink errors would be lost to the user, unless they add some handling themselves, right? Do akka-streams log that "Boom"?

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1008,4 +1009,99 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+            false
+          case t: T @unchecked =>
+            propagateOrElse(c2.sendOrClosed(t), other) {
+              propagateOrElse(other.sendOrClosed(t), c2)(true)
+            }
+      }
+    }
+    c2
+
+  private def propagateOrElse(unitOrClosed: Unit | ChannelClosed, otherSink: Sink[T])(ifNotClosed: => Boolean): Boolean =
```
adamw, line 1056: in fact, maybe this can be inline, together with the `ifNotClosed` parameter (removing the need for => then)?

adamw: but looks much better :)

core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -1008,4 +1009,99 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+    *
+    * If this source is failed then failure is passed to the returned and channel and the `other` Sink.
+    *
+    * @param other
+    *   The Sink to which elements from this source will be sent.
+    * @param queueSize
+    *   Specifies the size of the queue that buffers elements sent to the `other` Sink.
+    * @return
+    *   A source that emits input elements.
+    * @example
```
adamw, line 1075: I'd add a `@seeAlso` to `alsoTo` explaining the differences

core/src/main/scala/ox/channels/SourceOps.scala
adamw: They would not be lost, they would surface whenever you try to receive from the secondary sink

core/src/main/scala/ox/channels/SourceOps.scala
micossow: done

core/src/main/scala/ox/channels/SourceOps.scala
micossow: done

micossow:
To consider: as @adamw pointed out, because of `wireTap` working similarly to `alsoTo`, we should make the method names similar as well.
I was thinking about `alsoTo` and `alsoToIfAvailable` or sth.

adamw:
Maybe: `alsoToBoth` and `alsoToTap`? "both" stresses that values always go to both channels, "tap" signals that it's best-effort

kciesielski:
`alsoToTap` might be ok, but maybe let's keep `alsoTo` instead of `alsoToBoth`, sounds a bit off - it suggests that one sends "also to both other sinks".

micossow:
`alsoTo` and `alsoToTap` should do the job

core/src/test/scala/ox/channels/SourceOpsAlsoToTapTest.scala
```diff
@@ -0,0 +1,43 @@
+package ox.channels
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import ox.*
+import scala.concurrent.duration.*
+
+class SourceOpsAlsoToTapTest extends AnyFlatSpec with Matchers {
+
+  behavior of "Source.alsoToTap"
```
kciesielski, line 10: **suggestion:** Could you add a test to verify if failing `other` doesn't affect data flowing from the main result source?

kciesielski (approved):

PR #167, opened 2024-06-29
move catching to either object and make it an either block operator

lbialy:
also fixes unreachable case warning by adding specialised `ok()` operators:
```
[warn] -- [E030] Match case Unreachable Warning: /Users/lbialy/Projects/foss/ox/core/src/test/scala/ox/EitherTest.scala:145:65
[warn] 145 |    val r1 = either((1 to 20).toVector.mapPar(3)(i => Right(i).ok()))
[warn]     |                                                      ^^^^^^^^^^^^^
[warn]     |                                                      Unreachable case
[warn]     |---------------------------------------------------------------------------
[warn]     |Inline stack trace
[warn]     |- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
[warn]     |This location contains code that was inlined from either.scala:56
[warn]  56 |            case Left(e)  => break(Left(e))
[warn]     |                 ^^^^^^^
[warn]      ---------------------------------------------------------------------------
[warn] one warning found
```

### DIFF
--- a/core/src/main/scala/ox/either.scala
+++ b/core/src/main/scala/ox/either.scala
@@ -47,23 +47,6 @@ object either:
       ) nn: NotNested
   ): Either[E, A] = boundary(Right(body))
 
-  extension [E, A](inline t: Right[E, A])
-    /** Unwrap the value of the `Either`, returning value of type `A` on guaranteed `Right` case. */
-    transparent inline def ok(): A =
-      t match
-        case Right(a) => a
-
-  extension [E, A](inline t: Left[E, A])
-    /** Unwrap the value of the `Either`, short-circuiting the computation to the enclosing [[either]]. */
-    transparent inline def ok(): A =
-      summonFrom {
-        case given boundary.Label[Either[E, Nothing]] =>
-          break(t.asInstanceOf[Either[E, Nothing]])
-        case given boundary.Label[Either[Nothing, Nothing]] =>
-          error("The enclosing `either` call uses a different error type.\nIf it's explicitly typed, is the error type correct?")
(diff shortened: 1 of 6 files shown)


core/src/main/scala/ox/either.scala
```diff
@@ -42,6 +47,23 @@ object either:
       ) nn: NotNested
   ): Either[E, A] = boundary(Right(body))
 
+  extension [E, A](inline t: Right[E, A])
```
adamw, line 50: can you add the comment in code, as to why we need the specialised left/right variants? (avoiding compiler warnings)

core/src/main/scala/ox/either.scala
```diff
@@ -42,6 +47,23 @@ object either:
...
+    /** Unwrap the value of the `Either`, returning value of type `A` on guaranteed `Right` case. */
+    transparent inline def ok(): A =
```
adamw, line 52: this can be simplified to `t.value`?

core/src/main/scala/ox/either.scala
lbialy: sure

core/src/main/scala/ox/either.scala
lbialy: yeah!

core/src/main/scala/ox/either.scala
lbialy: done

core/src/main/scala/ox/either.scala
lbialy: done

adamw:
Thanks :)

PR #168, opened 2024-06-29
fix #164

### LINKED ISSUE #164: Two exceptions thrown at the same time cause the program to hang
This will print 1 and and then the InterruptedException in 2, while 3 will never show up and the whole application hangs.

```
@main def exceptionsOx() = 
  supervised:  
    def computation(withException: Option[String]): Int =
      sleep(2.seconds)
      withException match
        case None => 1
        case Some(value) =>
          throw new Exception(value)
    
    val fork1 = forkUser:
        computation(withException = None)
    val fork2 = forkUser:
        computation(withException = Some("Oh no!"))
    val fork3 = forkUser:
        computation(withException = Some("Oh well.."))
    println(fork1.joinEither()) // 1
    println(fork2.joinEither()) // 2
    println(fork3.joinEither()) // 3
```

This doesn't happen if we don't do a normal join. Is this an antipattern in ox? Looks like a very easy thing shot yourself with.

I mostly confused by the behaviour here.

lbialy:
Fork#joinEither aware whether fork finished with InterruptedException or not. Also adds sbt settings that greatly improve debugging of race conditions (because one can interrupt a forked jvm instead of killing sbt jvm).

### DIFF
--- a/core/src/main/scala/ox/fork.scala
+++ b/core/src/main/scala/ox/fork.scala
@@ -122,7 +122,7 @@ def forkAll[T](fs: Seq[() => T])(using Ox): Fork[Seq[T]] =
   val forks = fs.map(f => fork(f()))
   new Fork[Seq[T]]:
     override def join(): Seq[T] = forks.map(_.join())
-    override def wasInterrupted: Boolean = forks.forall(_.wasInterrupted)
+    override def wasInterruptedWith(ie: InterruptedException): Boolean = forks.exists(_.wasInterruptedWith(ie))
 
 /** Starts a fork (logical thread of execution), which is guaranteed to complete before the enclosing [[supervised]], [[supervisedError]] or
   * [[unsupervised]] block completes, and which can be cancelled on-demand.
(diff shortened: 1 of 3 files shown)


core/src/main/scala/ox/fork.scala
```diff
@@ -177,8 +178,13 @@ def forkCancellable[T](f: => T)(using OxUnsupervised): CancellableFork[T] =
       if !started.getAndSet(true)
       then result.completeExceptionally(new InterruptedException("fork was cancelled before it started")).discard
 
+    override def wasInterruptedWith(ie: InterruptedException): Boolean =
+      result.isCompletedExceptionally && (result.exceptionNow() eq ie)
```
hearnadam, line 182: Does this need to be public? it seems like it has the potential to be misused given the `eq` check.

lbialy:
Nope, let me push the commit :)

core/src/main/scala/ox/fork.scala
lbialy: yeah, we discussed that in the issue, it's package private now.

adamw:
Thanks! :)

PR #172, opened 2024-07-03
Scheduled and repeat API + retry refactoring

### LINKED ISSUE #57: Add a schedule-aware repeat
While the retry mechanism allows for repeating an operation - according to a `Schedule` in case of error, we'd also like to have a schedule-aware `repeat`, which would run an operation according to a `Schedule` or perhaps until a stop condition is met.

Some things to consider:
- what happens if the operation fails before the scheduled number of repetitions? Maybe this could be expressed with a multi-policy (see: #55) that combines `repeat` and `retry`?
- perhaps the stop condition could be thought of yet another `Schedule` like `UntilResult[T](p: T => Boolean)` or `Until(p: => Boolean)`, or both?

micossow:
closes #57 

Summary:
* `retry*` API has become more generic and renamed to `scheduled*`
* existing `retry*` API uses `scheduled*` internally and retry-specific DSL has remained unchanged
* new `repeat*` API also uses scheduled*` internally with repeat-specific DSL
* `scheduled` API, `Schedule` trait and its implementations should be decoupled from `retry` API, specifically it calculates "duration" instead of "delay", which can later be interpreted as a delay or interval.
* `scheduled` implementation is covered by tests for `retry` and `repeat`
* docs updated where applicable


core/src/main/scala/ox/resilience/retry.scala
```diff
@@ -48,33 +47,12 @@ def retryEither[E, T](policy: RetryPolicy[E, T])(operation: => Either[E, T]): Ei
...
-    case _                               => None
-
-  loop(1, remainingAttempts, None)
+  runScheduledWithErrorMode(em)(
+    RunScheduledConfig(
+      policy.schedule,
+      policy.onRetry,
+      shouldContinueOnError = policy.resultPolicy.isWorthRetrying,
+      shouldContinueOnResult = t => !policy.resultPolicy.isSuccess(t),
+      delayPolicy = DelayPolicy.SinceTheEndOfTheLastInvocation
```
adamw, line 56: shouldn't the delay policy be part of `RetryPolicy`, so that it's also user-configurable if needed?

core/src/main/scala/ox/scheduling/RepeatConfig.scala
```diff
@@ -0,0 +1,30 @@
+package ox.scheduling
+
+import scala.concurrent.duration.FiniteDuration
+import scala.concurrent.duration.DurationInt
+
+case class RepeatConfig[E, T](
```
adamw, line 6: we have `RepeatConfig`, but `RetryPolicy` - why?

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -0,0 +1,167 @@
...
+sealed trait Schedule:
+  // TODO: consider better name that `attempt`
+  def nextDelay(attempt: Int, lastDelay: Option[FiniteDuration]): FiniteDuration
+
+object Schedule:
+
+  private[scheduling] sealed trait Finite extends Schedule:
+    def maxRepeats: Int
+    def initialDelay: FiniteDuration = Duration.Zero
+    def fallbackTo(fallback: Finite): Finite = FallingBack(this, fallback)
```
adamw, line 16: maybe `.andThen` would be a better name?

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -0,0 +1,167 @@
+package ox.scheduling
+
+import scala.concurrent.duration.*
+import scala.util.Random
+
+// TODO: shouldn't this trait be unsealed so that users can create their own schedules?
+sealed trait Schedule:
+  // TODO: consider better name that `attempt`
+  def nextDelay(attempt: Int, lastDelay: Option[FiniteDuration]): FiniteDuration
```
adamw, line 9: attempt - that's the attempt number, yes?

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -0,0 +1,167 @@
+package ox.scheduling
+
+import scala.concurrent.duration.*
+import scala.util.Random
+
+// TODO: shouldn't this trait be unsealed so that users can create their own schedules?
+sealed trait Schedule:
```
adamw, line 7: let's keep it sealed for now, and wait for a use-case to unseal it ;)

core/src/main/scala/ox/scheduling/runScheduled.scala
```diff
@@ -0,0 +1,67 @@
...
+import ox.{EitherMode, ErrorMode, sleep}
+
+import scala.annotation.tailrec
+import scala.concurrent.duration.{Duration, FiniteDuration, DurationLong}
+import scala.util.Try
+
+enum DelayPolicy:
+  case SinceTheStartOfTheLastInvocation, SinceTheEndOfTheLastInvocation
+
+case class RunScheduledConfig[E, T](
```
adamw, line 12: these classes/methods can all be `private[ox]` I suppose? they're not meant to be used by the end-user?

adamw:
Looks good but it would be great to see some examples of the API in action (in docs, or stand-alone) to see how the API can be used

core/src/main/scala/ox/resilience/retry.scala
micossow: My idea was to create a public API `runScheduled*` with the all configurable options and make `retry*` and `repeat*` as "thin" as possible - basically an opinionated variants of `runScheduled*`, specifically:
* `retry` - that behaves similarly to resillience4j retries, where sleep time is counted since the end of previous operation, has an optional `onRetry` callback and initial delay (sleep before first operation) is not relevant
* `repeat` - that behaves similarly to Akka's `Source.tick`, where sleep time is counted since the start of previous operation (interval), doesn't expose `onRepeat` callback (not relevant) and initial delay is allowed
WDYT?

adamw: Ok, sounds good. Let's just document these defaults quite clearly, plus all the flexibility that `runScheduled` (or maybe simply `scheduled`?) gives you

core/src/main/scala/ox/scheduling/RepeatConfig.scala
micossow: `RetryPolicy` has been refactored to `RetryConfig`

core/src/main/scala/ox/scheduling/Schedule.scala
micossow: done

core/src/main/scala/ox/scheduling/Schedule.scala
micossow: For retries it is an attempt number, for repeats it's an repetition number. I'm still looking for some common name :)

core/src/main/scala/ox/scheduling/Schedule.scala
micossow: todo removed

core/src/main/scala/ox/scheduling/runScheduled.scala
micossow: it has been decided that the `scheduled` and `ScheduledConfig` will be public API

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -0,0 +1,166 @@
+package ox.scheduling
+
+import scala.concurrent.duration.*
+import scala.util.Random
+
+sealed trait Schedule:
+  // TODO: consider better name that `attempt`
```
adamw, line 7: iteration?

core/src/main/scala/ox/scheduling/scheduled.scala
```diff
@@ -0,0 +1,130 @@
...
+import ox.{EitherMode, ErrorMode, sleep}
+
+import scala.annotation.tailrec
+import scala.concurrent.duration.{Duration, FiniteDuration, DurationLong}
+import scala.util.Try
+
+/** The mode that specifies how to interpret the duration provided by the schedule. */
+enum SleepMode:
+  /** Interval (since the start of the last operation), i.e. guarantees that the next operation will start no sooner than the specified
+    * duration after the previous operations has finished. If the previous operation takes longer than the interval, the next operation will
```
adamw, line 12: started, not finished?

doc/adr/0008-scheduled-repeat-retry.md
```diff
@@ -0,0 +1,18 @@
+# 8. Retries
```
adamw, line 1: Good idea to add an ADR :)

doc/repeat.md
```diff
@@ -0,0 +1,71 @@
...
+* directly using a by-name parameter, i.e. `f: => T`
+* using a by-name `Either[E, T]`
+* or using an arbitrary [error mode](basics/error-handling.md), accepting the computation in an `F` context: `f: => F[T]`.
+
+## Configuration
+
+The `repeat` config requires a `Schedule`, which indicates how many times and with what interval should the `operation` 
+be repeated.
+
+In addition, it is possible to define strategies for handling the results and errors returned by the `operation`:
```
kciesielski, line 30: **question:** Shouldn't we have a stronger division between `retry`, which is related to "trying and possibly failing" and `repeat`, which has no notion of errors? I think it would be simpler from the user perspective, otherwise they may be confused when to use retry and when to use repeat + `shouldContinueOnError`.

doc/scheduled.md
```diff
@@ -0,0 +1,71 @@
...
+
+Similarly to the `retry` and `repeat` APIs, the `operation` can be defined: 
+* directly using a by-name parameter, i.e. `f: => T`
+* using a by-name `Either[E, T]`
+* or using an arbitrary [error mode](basics/error-handling.md), accepting the computation in an `F` context: `f: => F[T]`.
+
+## Configuration
+
+The `scheduled` config consists of:
+- a `Schedule`, which indicates how many times and with what delay/interval should the `operation`be run.
```
kciesielski, line 16: **issue:** The `Schedule` indicates what is the delay/interval. Then `SleepMode` again defines an `Interval` or a `Delay` - sounds confusing?
Also, regarding the distinction between `interval` and `delay` - I guess they correspond to `fixedDelay` and `fixedRate` from Akka, right? Maybe such naming would be more clear. For example `FixedDelay` and `FixedRateInterval`?

micossow: I wanted to make the `Schedule` trait repeat/retry agnostic, that's why it provides `nextDuration`, not `nextDelay` nor `nextInterval` and user can decide later when using `scheduled` how that duration will be interpreted.

doc/repeat.md
micossow: Right, I don't think that `shouldContinueOnError` handler would be used often.

core/src/main/scala/ox/scheduling/Schedule.scala
micossow: Yes, I think it would be the best in that case

core/src/main/scala/ox/scheduling/scheduled.scala
micossow: fixed

doc/scheduled.md
kciesielski: Ok, but regarding my first question - docs suggest that both the `Schedule` and `SleepMode` are responsible for the delay/interval.

core/src/main/scala/ox/resilience/retry.scala
```diff
@@ -1,80 +1,50 @@
...
-import ox.{EitherMode, ErrorMode, sleep}
+import ox.{EitherMode, ErrorMode}
+import ox.scheduling.*
 
-import scala.annotation.tailrec
-import scala.concurrent.duration.*
 import scala.util.Try
 
-/** Retries an operation returning a direct result until it succeeds or the policy decides to stop.
+/** Retries an operation returning a direct result until it succeeds or the config decides to stop.
```
adamw, line 8: I think I'd also mention here that `retry` is a special-case of `schedule`, with a given set of defaults (and enumarte these defaults). Maybe this should be in the `@see` section?

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -0,0 +1,165 @@
+package ox.scheduling
+
+import scala.concurrent.duration.*
+import scala.util.Random
+
+sealed trait Schedule:
+  def nextDuration(iteration: Int, lastDuration: Option[FiniteDuration]): FiniteDuration
```
adamw, line 7: or ... we're using `invocation` in the docs quite often, so maybe that would be a good name as well?

core/src/main/scala/ox/resilience/retry.scala
micossow: done

doc/repeat.md
micossow: `shouldContinueOnError` removed and mentioned that `repeat` can be used with `retry` + example

core/src/main/scala/ox/scheduling/Schedule.scala
micossow: good idea!

doc/repeat.md
```diff
@@ -0,0 +1,78 @@
...
+
+### API shorthands
+
+You can use one of the following shorthands to define a `RepeatConfig` with a given schedule with an optional initial delay:
+- `RepeatConfig.immediate(maxInvocations: Int, initialDelay: Option[FiniteDuration] = None)`
+- `RepeatConfig.immediateForever[E, T](initialDelay: Option[FiniteDuration] = None)`
+- `RepeatConfig.fixedRate[E, T](maxInvocations: Int, interval: FiniteDuration, initialDelay: Option[FiniteDuration] = None)`
+- `RepeatConfig.fixedRateForever[E, T](interval: FiniteDuration, initialDelay: Option[FiniteDuration] = None)`
+- `RepeatConfig.backoff[E, T](maxInvocations: Int, firstInterval: FiniteDuration, maxInterval: FiniteDuration = 1.minute, jitter: Jitter = Jitter.None, initialDelay: Option[FiniteDuration] = None)`
+- `RepeatConfig.backoffForever[E, T](firstInterval: FiniteDuration, maxInterval: FiniteDuration = 1.minute, jitter: Jitter = Jitter.None, initialDelay: Option[FiniteDuration] = None)`
```
adamw, line 45: it would be good to give an example of combinining schedules using `.andThen` (here and in the code snippet)

adamw (approved):
Just some minor adjustments, but looks great otherwise!

doc/repeat.md
micossow: added some examples to `scheduled.md` and `See [scheduled](scheduled.md) for details on how to create custom schedules.` to `repeat.md` and `retries.md`

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -0,0 +1,165 @@
...
+    * [[second.maxRepeats]] number of times.
+    */
+  private[scheduling] case class FiniteAndFiniteSchedules(first: Finite, second: Finite) extends CombinedSchedules, Finite:
+    override def maxRepeats: Int = first.maxRepeats + second.maxRepeats
+    override def initialDelay: FiniteDuration = first.initialDelay
+
+  private[scheduling] object FiniteAndFiniteSchedules:
+    /** A schedule that repeats indefinitely, using [[first]] first [[first.maxRepeats]] number of times, and then always using [[second]].
+      */
+    def forever(first: Finite, second: Infinite): Infinite = FiniteAndInfiniteSchedules(first, second)
```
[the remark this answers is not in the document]
micossow: yes, it doesn't make sense after refactoring. Fixed!

doc/scheduled.md
kciesielski: Crystal clear :gem:, thanks!

kciesielski (approved):

PR #179, opened 2024-07-11
Log exceptions that occur during shutdown, if they are not part of the "normal" interruption process

### LINKED ISSUE #176: finalisers swallow exceptions
So it seems exceptions are swallowed by finalisers such as `releaseAfterScope`

```scala
//> using scala 3.4.2
//> using jvm 21
//> using dep "com.softwaremill.ox::core:0.3.0"

import ox.*

object launch extends OxApp.Simple:
  def run(using Ox): Unit =
    releaseAfterScope:
      println("cancelled")
      ??? // (here!)
    never
```

if I cancel this app, it will print "cancelled", but there is no mention of the exception:
```
scala run .
^Ccancelled
```

I imagine that it should be useful to report on any exception thrown (except for control flow such as interruption)

## Discussion in https://github.com/softwaremill/ox/issues/174

So the problem is with exceptions being thrown from finalizers? Where would you expect to see this exception?

Not sure how to handle those ... when there's an interrupt, all forks that are running will end up with an exception. Most of these should be `InterruptedException` - I guess these should be ignored?

Maybe we should have another (overridable, or as part of `OxAppConfiguration`) method which is called for the exception with which the main scope ends, and then it could by default properly report any that are **not** InterruptedExceptions?

_Originally posted by @adamw in https://github.com/softwaremill/ox/issues/174#issuecomment-2218669757_

adamw:
Closes #176

Also introduces configurable exception handling in the main app body.

### DIFF
--- a/core/src/main/scala/ox/OxApp.scala
+++ b/core/src/main/scala/ox/OxApp.scala
@@ -16,8 +16,11 @@ enum ExitCode(val code: Int):
   *   - [[OxApp.WithErrorMode]] to report errors (which end the application) using other [[ErrorMode]]s
   *
   * The benefit of using `OxApp` compared to normal `@main` methods is that application interruptions is handled properly. A fork in a scope
-  * is created to run the application's logic. Interrupting the application (e.g. using CTRL+C) will cause the scope to end, resources to be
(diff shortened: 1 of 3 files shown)


bishabosha:
amazing

Issue #197, opened 2024-08-18
`flatten` method for `Stream[Stream[T]]`


nimatrueway:
Is there any tool of such nature for `Stream`, I was writing a few examples for my company to compare different Scala streaming libraries (e.g. monix `Observable`, fs2 `Stream`, ox `Source`, etc.) that I run into this case.

```scala
Source.fromIterable(splitRange(request.startDate, request.endDate)) // = STREAM[(Instant, Instant)]
  .mapPar(4) { case (start, end) =>
    service.fetch(start, end) // = STREAM[T]
  } // = STREAM[STREAM[T]]
  .flatten // DOES NOT EXIST 😭
```

I was just wondering if this is overlooked or is there any solid reason as to not include `flatten` in `Source` api?

nimatrueway:
This weekend I cludged this together, I can send up a pull-request 🤔

```
  extension [T](source: Source[Source[T]]) {
    def flatten(using Ox, StageCapacity): Source[T] = {
      val c2 = StageCapacity.newChannel[T]
      val closeScope = Phaser(1) // waitlist (output stream plus all child streams)

      forkPropagate(c2) {
        repeatWhile {
          source.receiveOrClosed() match
            case ChannelClosed.Done =>
              closeScope.arriveAndAwaitAdvance()
              c2.doneOrClosed()
              false
            case ChannelClosed.Error(r) =>
              c2.errorOrClosed(r)
              false
            case t: Source[T] @unchecked =>
              closeScope.register()
              // asynchronously pipe the child stream elements to the output stream
              forkPropagate(c2) {
                try
                  repeatWhile {
                    t.receiveOrClosed() match
                      case ChannelClosed.Done     => false
                      case e: ChannelClosed.Error => c2.errorOrClosed(e.reason); false
                      case t: T @unchecked        => c2.send(t); true
                  }
                finally
                  closeScope.arriveAndDeregister()
              }
              true
        }
      }
      c2
    }
  }
```

nimatrueway:
I looked into how this can be implemented, I suspect that we need a `select` that it allows a mutable list of `ReceiveClause` so that we can add to the clauses as the parent source generates more children sources 🤔

PR #198, opened 2024-08-19
Add `flatten` to Source[T]

nimatrueway:
To address https://github.com/softwaremill/ox/issues/197


core/src/main/scala/ox/channels/SourceOps.scala
```diff
@@ -393,6 +393,39 @@ trait SourceOps[+T] { outer: Source[T] =>
...
+  /** Pipes the elements of child sources into the output source. If the parent source or any of the child sources emit an error, the
+    * pulling stops and the output source emits the error.
+    */
+  def flatten[U](using Ox, StageCapacity, T <:< Source[U]): Source[U] = {
+    val c2 = StageCapacity.newChannel[U]
+
+    forkPropagate(c2) {
+      var pool = List[Source[T] | Source[U]](this)
+      repeatWhile {
+        selectOrClosed(pool) match {
```
nimatrueway, line 405: The fact that we can not know the result of select stems from which select statement is limiting. In Golang there's a hacky way to do it.
https://stackoverflow.com/a/19992525/1556045

I think both `(v: V)` or `(c: ChannelClosed)` results if they hint that which select/source they stem from would make the API more robust.

adamw:
Thanks for the PR - the only reason such a method didn't exist before is because nobody wrote it :). There are some test failures, though.

As for adding a select variant which would also return the clause which was completed - maybe you can create a separate issue for that?

nimatrueway:
> Thanks for the PR - the only reason such a method didn't exist before is because nobody wrote it :). There are some test failures, though.
> 
> As for adding a select variant which would also return the clause which was completed - maybe you can create a separate issue for that?

Thanks for taking a look @adamw, I created a new ticket and made the test pass. Let me know if there's any concern with the readability of the tests or more test cases would help.

adamw:
The test are great, thanks! :)

PR #216, opened 2024-09-16
Handle ControlThrowable exception in 'race' function

### LINKED ISSUE #213: The `race` function doesn't handle functions that can throw a `ControlThrowable` exception
I'm the maintainer of the library `raise4s`(link to the repo [here](https://github.com/rcardin/raise4s)). I use context functions to let the user declare functions that can raise logic-typed errors instead of exceptions. Jon Pretty is doing something close to my library with [Contingency](https://soundness.dev/contingency/).

I'd love to integrate with `ox` and write code like the following:

```scala 3
@main
def main(): Unit = {

  val result: Int raises String = ox.race(
    {
      ox.sleep(500.milliseconds)
      println("short task")
      Raise.raise("Something went wrong")
      2
    }, {
      ox.sleep(1.second)
      println("long task")
      Raise.raise("Something went wrong too late")
      1
    }
  )

  Raise.run(result) match {
    case value: Int => println(s"Value: $value")
    case error: String => println(s"Error: $error")
  }
}
```

However, the above code doesn't work. In detail, the execution of the `race` function never stops.

The implementation of the `raise` method uses the following exception:

```scala 3
private[raise4s] case class Raised[Error](original: Error)
    extends ControlThrowable
    with NoStackTrace
```

However, the implementation of the `ox.race` function handles the execution of the given lambdas through a `Try`:

```scala 3
fs.foreach(f => forkUnsupervised(result.put(Try(f()))))
```

Unfortunately, the `Try` type doesn't handle exceptions extending `ControlThrowable`. You can find [here](https://contributors.scala-lang.org/t/nonfatal-and-controlthrowable-changed-in-stdlib/6049/1) more details.

I already tried changing the `ox` code locally to add the handling of `ControlThrowable,` and everything started to integrate smoothly.

Can I open a PR to address the problem? Would you like to chat about it first?

rcardin:
I added the handling of a `ControThrowable` to the internals of the `race` function. The `ControThrowable` exceptions have suppression disabled by default, so it's not possible to accumulate suppressed exceptions.

Feel free to let me know if anything needs clarification or if you notice something needs to be improved.

Closes #213

### DIFF
--- a/core/src/main/scala/ox/race.scala
+++ b/core/src/main/scala/ox/race.scala
@@ -4,6 +4,7 @@ import java.util.concurrent.ArrayBlockingQueue
 import scala.annotation.tailrec
 import scala.concurrent.TimeoutException
 import scala.concurrent.duration.FiniteDuration
+import scala.util.control.{ControlThrowable, NonFatal}
 import scala.util.{Failure, Success, Try}
 
 /** A `Some` if the computation `t` took less than `duration`, and `None` otherwise. if the computation `t` throws an exception, it is
@@ -38,7 +39,13 @@ def race[T](fs: Seq[() => T]): T = race(NoErrorMode)(fs)
 def race[E, F[_], T](em: ErrorMode[E, F])(fs: Seq[() => F[T]]): F[T] =
   unsupervised {
     val result = new ArrayBlockingQueue[Try[F[T]]](fs.size)
-    fs.foreach(f => forkUnsupervised(result.put(Try(f()))))
+    fs.foreach(f => forkUnsupervised {
+      val r = try Success(f())
+      catch 
+        case NonFatal(e) => Failure(e)
+        case e: ControlThrowable => Failure(e)
+      result.put(r)
+    })
 
     @tailrec
     def takeUntilSuccess(failures: Vector[Either[E, Throwable]], left: Int): F[T] =
--- a/core/src/main/scala/ox/race.scala
+++ b/core/src/main/scala/ox/race.scala
@@ -38,14 +38,22 @@ def race[T](fs: Seq[() => T]): T = race(NoErrorMode)(fs)
   */
 def race[E, F[_], T](em: ErrorMode[E, F])(fs: Seq[() => F[T]]): F[T] =
   unsupervised {
-    val result = new ArrayBlockingQueue[Try[F[T]]](fs.size)
-    fs.foreach(f => forkUnsupervised {
-      val r = try Success(f())
-      catch 
-        case NonFatal(e) => Failure(e)
-        case e: ControlThrowable => Failure(e)
-      result.put(r)
-    })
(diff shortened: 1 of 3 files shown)


core/src/main/scala/ox/race.scala
```diff
@@ -38,7 +39,13 @@ def race[T](fs: Seq[() => T]): T = race(NoErrorMode)(fs)
 def race[E, F[_], T](em: ErrorMode[E, F])(fs: Seq[() => F[T]]): F[T] =
   unsupervised {
     val result = new ArrayBlockingQueue[Try[F[T]]](fs.size)
-    fs.foreach(f => forkUnsupervised(result.put(Try(f()))))
+    fs.foreach(f => forkUnsupervised {
+      val r = try Success(f())
+      catch 
+        case NonFatal(e) => Failure(e)
+        case e: ControlThrowable => Failure(e)
```
adamw, line 46: I'm actually wondering if we shouldn't introduce a third state here. So `race` (as opposed to `raceResult`) returns the first success, or returns a failure if all computations fail. So the question is, should we also wait for other computations to fail, if one of the reports throws a `ControlThrowable`? Maybe you can share a use-case?

I was thinking about the other exceptions that might get thrown. `InterruptedException`, `VirtualMachineError`, `ThreadDeath`, `LinkageError` are all exceptions which should cause `race` to finish immediately (without checking if other branches throw exceptions, or not). There's still a bug in the current implementation as to how these are handled, but I'm not 100% sure how to handle thee fatal exceptions in terms of the contract of `race`.

rcardin: `Try` doesn’t handle fatal exceptions. Do you think we need to handle also Fatal?

rcardin: > Maybe you can share a use-case?
Sure. `raise4s` uses a `ControlThrowable` to forward typed errors to the caller, as they were exceptions. So, take the following code for example:

```scala 3
val result: Int raises String = ox.race(
  {
    ox.sleep(500.milliseconds)
    Raise.raise("Something went wrong")
  }, {
    ox.sleep(1.second)
    1
  }
)
Raise.run(result) match {
  case value: Int => println(s"Value: $value")
  case error: String => println(s"Error: $error")
}
```

I expected the above code to print `Value: 1` to the console, waiting for the long task to finish, since the first went in a logic error. I don't know if we should make this behavior universal for `ControlThrowable`s.

>  but I'm not 100% sure how to handle the fatal exceptions in terms of the contract of `race`

AFAIK the `Fatal` exception should not be handled. I agree that if the code throws a fatal exception, the whole `race` must stop immediately and rethrow the fatal error.

Please let me know your thoughts and how you'd like me to proceed.

adamw:
Thanks, I've extended the code to also handle other fatal exceptions

PR #251, opened 2024-12-09
RateLimiter - consider whole operation execution time

### LINKED ISSUE #246: Extend rate limiter to support whole operation execution time, not just the start
Currently, the rate limiter only takes into account the start of execution of an operation - and limits basing on this.

However, in some scenarios we should take into account the whole duration. E.g. if there's an HTTP request which we want to rate limit, the actual HTTP invocation might happen anytime between the start and end of an `httpRequest.send(...)` call.

The goal of this task is to design an API which would extend what's currently available, and support both start-time & whole-duration rate limits, using all three algorithms (fixed windows, sliding window, leaky bucket)

Kamil-Lontkowski:
Closes #246


core/src/main/scala/ox/resilience/RateLimiter.scala
```diff
@@ -3,28 +3,47 @@ package ox.resilience
...
+/** Rate limiter with a customizable algorithm. Operations can be blocked or dropped, when the rate limit is reached. considerOperationTime decides if whole time of execution should be considered or just the start. */
+class RateLimiter private (algorithm: RateLimiterAlgorithm, considerOperationTime: Boolean):
+  private val semaphore = new Semaphore(algorithm.rate)
+
   /** Runs the operation, blocking if the rate limit is reached, until the rate limiter is replenished. */
   def runBlocking[T](operation: => T): T =
-    algorithm.acquire()
-    operation
+    if considerOperationTime then
+      synchronized:
```
adamw, line 16: we can't (and shouldn't) use `synchronized` in Loom (at least until JDK25). Besides, all synchronization is done using semaphores, so this shouldn't be necessary

core/src/main/scala/ox/resilience/RateLimiter.scala
```diff
@@ -3,28 +3,47 @@ package ox.resilience
 import scala.concurrent.duration.FiniteDuration
 import ox.*
 
+import java.util.concurrent.Semaphore
 import scala.annotation.tailrec
 
-/** Rate limiter with a customizable algorithm. Operations can be blocked or dropped, when the rate limit is reached. */
-class RateLimiter private (algorithm: RateLimiterAlgorithm):
+/** Rate limiter with a customizable algorithm. Operations can be blocked or dropped, when the rate limit is reached. considerOperationTime decides if whole time of execution should be considered or just the start. */
+class RateLimiter private (algorithm: RateLimiterAlgorithm, considerOperationTime: Boolean):
```
adamw, line 10: maye instead of a boolean flag we could have an enum, sth like `RateLimiterMode` with `OperationStart` and `OperationDuration` values. This should be clearer for users

core/src/main/scala/ox/resilience/RateLimiter.scala
```diff
@@ -3,28 +3,47 @@ package ox.resilience
...
+      operation
 
   /** Runs or drops the operation, if the rate limit is reached.
     *
     * @return
     *   `Some` if the operation has been allowed to run, `None` if the operation has been dropped.
     */
   def runOrDrop[T](operation: => T): Option[T] =
-    if algorithm.tryAcquire() then Some(operation)
+    if considerOperationTime then
```
adamw, line 32: an option to explore: instead of if-else, maybe the operation-time rate limiters should be another `Algorithm` implementations? (with some extensions to the algorithm interface)

core/src/main/scala/ox/resilience/RateLimiter.scala
```diff
@@ -72,7 +95,9 @@ object RateLimiter:
     *   Max capacity of tokens in the algorithm, limiting the operations that are allowed to **start** concurrently.
     * @param refillInterval
     *   Interval of time between adding a single token to the bucket.
+    * @param considerOperationTime
+    *   Whether to consider whole execution time of operation or just the start.
     */
-  def leakyBucket(maxTokens: Int, refillInterval: FiniteDuration)(using Ox): RateLimiter =
-    apply(RateLimiterAlgorithm.LeakyBucket(maxTokens, refillInterval))
+  def leakyBucket(maxTokens: Int, refillInterval: FiniteDuration, considerOperationTime: Boolean = false)(using Ox): RateLimiter =
```
adamw, line 101: does leaky bucket make sense with whole-operation-time mode?

Kamil-Lontkowski: In leaky bucket we have `Semaphore(1)` and it's released every `per`(if it lower than rate). So we still can have case where we submit long operation, then the bucket refill and let us submit another operation while first is still running. At least I understand it that way looking at `LeakyBucket` code.

core/src/main/scala/ox/resilience/RateLimiter.scala
Kamil-Lontkowski: I started with that option, but in the end it produced a duplicated algorithms with logic that is now in `runOrDrop` and `runBlocking` in different places. Currently when we want to track execution time we have two semaphores that can block independently, so for now I don't see a reason to move this logic to `RateLimiterAlgorithm`, maybe with different algorithms it would need to be done this way, but for now I think this is better option.

core/src/main/scala/ox/resilience/RateLimiter.scala
```diff
@@ -3,28 +3,52 @@ package ox.resilience
...
+import java.util.concurrent.Semaphore
 import scala.annotation.tailrec
 
-/** Rate limiter with a customizable algorithm. Operations can be blocked or dropped, when the rate limit is reached. */
-class RateLimiter private (algorithm: RateLimiterAlgorithm):
+/** Rate limiter with a customizable algorithm. Operations can be blocked or dropped, when the rate limit is reached. operationMode decides
+  * if whole time of execution should be considered or just the start.
+  */
+class RateLimiter private (algorithm: RateLimiterAlgorithm, operationMode: RateLimiterMode):
+  private val semaphore = new Semaphore(algorithm.rate)
```
adamw, line 13: I don't think that's a correct solution. If I understand this correctly, the main change is adding a semaphore which causes that at most `rate` computations run concurrently. But rate limiting is about time, which is not present here at all.

Suppose we use the fixed-window rate limiter, with a limit of 2 ops/second. Consider this scenario:
1. we start two operations at t=900ms; they will take 200ms
2. at t=1000ms, we replenish the permits (as the algorithm specifies)
3. at t=1100ms, the two operations finish, and we release the `semaphore`
4. at t>1100ms, we can start a new operations, but we shouldn't be able to - with `OperationDuration` we should consider the whole operation's time, including the fact that it spanned two windows, so both of these might as well have actually happened in the second time window

For the next iteration, please also provide an explanation of your approach, and why it works

core/src/main/scala/ox/resilience/RateLimiter.scala
```diff
@@ -46,11 +66,15 @@ object RateLimiter:
     * @param maxOperations
     *   Maximum number of operations that are allowed to **start** within a time [[window]].
     * @param window
-    *   Interval of time between replenishing the rate limiter. THe rate limiter is replenished to allow up to [[maxOperations]] in the next
+    *   Interval of time between replenishing the rate limiter. The rate limiter is replenished to allow up to [[maxOperations]] in the next
     *   time window.
     */
-  def fixedWindow(maxOperations: Int, window: FiniteDuration)(using Ox): RateLimiter =
-    apply(RateLimiterAlgorithm.FixedWindow(maxOperations, window))
+  def fixedWindow(maxOperations: Int, window: FiniteDuration, operationMode: RateLimiterMode = RateLimiterMode.OperationStart)(using
```
uini223, line 72: I'm not sure if we need those extensions here. 
Overall we have DurationRateLimiter with factory methods and imho it should be enough. 

We don't want to change the behaviour of those methods anyway. 

I would additionally add info in the docs that this is operation start based and for duration based see other.

uini223 (changes requested):

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
```diff
@@ -0,0 +1,158 @@
...
+      ()
+
+    def update(): Unit =
+      val now = System.nanoTime()
+      // retrieving current queue to append it later if some elements were added concurrently
+      val q = log.getAndUpdate(_ => Queue[(Long, Int)]())
+      // remove records older than window size
+      val qUpdated = removeRecords(q, now)
+      // merge old records with the ones concurrently added
+      val _ = log.updateAndGet(qNew =>
```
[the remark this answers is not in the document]
Kamil-Lontkowski: This is still kind of WIP, but this was copied from RateLimiterAlgorithm. I guess it was just another way for the last expression to be Unit. I will refactor this to be consistent.

core/src/main/scala/ox/resilience/RateLimiter.scala
```diff
@@ -3,23 +3,43 @@ package ox.resilience
...
+/** Rate limiter with a customizable algorithm. Operations can be blocked or dropped, when the rate limit is reached. operationMode decides
+  * if whole time of execution should be considered or just the start.
+  */
 class RateLimiter private (algorithm: RateLimiterAlgorithm):
   /** Runs the operation, blocking if the rate limit is reached, until the rate limiter is replenished. */
   def runBlocking[T](operation: => T): T =
-    algorithm.acquire()
-    operation
+    algorithm match
+      case alg: DurationRateLimiterAlgorithm =>
```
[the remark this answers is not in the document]
Kamil-Lontkowski: Yes, I was having trouble with coming up with sensible interface. I think that'll be better solution.

core/src/main/scala/ox/resilience/RateLimiter.scala
Kamil-Lontkowski: According to previous comment. If we just change RateLimiter and use:
```scala
algorithm.acquire()
alogrithm.runOperation(operation)
```
I don't think we then need DurationRateLimiter, we just change the algorithm provided to RateLimiter. If we go this route I think extensions that let you define RateLimter with DurationRateLimiterAlgorithm would be useful. Maybe new methods instead of extending existing ones.

Kamil-Lontkowski:
Considering [comment](https://github.com/softwaremill/ox/pull/251#discussion_r1878249002).

`DurationRateLimiterAlgorithm.FixedWindow` tracks number of running operations and doesn't replenish number of permits that running operations are "blocking".

`DurationRateLimiterAlgorithm.SlidingWindow` adds entry to log only after the operation is finished. This means that we consider end of operation as a point where we can replenish permits after `per` passes.

These changes mean that fixed window behaves in example as specified in [the comment](https://github.com/softwaremill/ox/pull/251#discussion_r1878249002]).

For sliding window consider this example:
We have three types of operations - instant, short(300 millis), long (3 seconds)
```scala
val rateLimiter = RateLimiter.slidingWindow(3, FiniteDuration(1, "second"), OperationDuration)

forkUserDiscard:
  result1 = rateLimiter.runOrDrop(shortOperation)
forkUserDiscard:
  result2 = rateLimiter.runOrDrop(shortOperation)
forkUserDiscard:
  result3 = rateLimiter.runOrDrop(longOperation)
forkUserDiscard:
  sleep(1500.millis)
  result4 = rateLimiter.runOrDrop(instantOperation)
  result5 = rateLimiter.runOrDrop(instantOperation)
  result6 = rateLimiter.runOrDrop(instantOperation)
```
After 1.5 second, two of three operations ended(at 0.3 seconds, so one full second passed since then), so we should have 2 available permits. Since at this point long operation is still running, we drop last operation and `result6` is None

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
```diff
@@ -0,0 +1,108 @@
...
+      semaphore.tryAcquire(permits)
+
+    def getNextUpdate: Long =
+      val waitTime = lastUpdate.get() + per.toNanos - System.nanoTime()
+      if waitTime > 0 then waitTime else 0L
+
+    def update(): Unit =
+      val now = System.nanoTime()
+      lastUpdate.set(now)
+      semaphore.release(rate - semaphore.availablePermits() - runningOperations.get())
```
adamw, line 31: let's add a comment here, why we calculate the permits to release this way

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
```diff
@@ -0,0 +1,108 @@
...
+    end update
+
+    def runOperation[T](operation: => T, permits: Int): T =
+      runningOperations.updateAndGet(_ + permits)
+      val result = operation
+      runningOperations.updateAndGet(current => (current - permits).max(0))
```
adamw, line 37: I guess this should never fall below 0? and if it does, that's a bug? if so, maybe let's add an assert here

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
```diff
@@ -0,0 +1,108 @@
...
+
+import java.util.concurrent.Semaphore
+import java.util.concurrent.atomic.AtomicInteger
+import java.util.concurrent.atomic.AtomicLong
+import java.util.concurrent.atomic.AtomicReference
+import scala.annotation.tailrec
+import scala.collection.immutable.Queue
+import scala.concurrent.duration.FiniteDuration
+
+object DurationRateLimiterAlgorithm:
```
adamw, line 11: lets add a comment here describing how this differs from `RateLimiterAlgorithm`

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
```diff
@@ -0,0 +1,108 @@
...
+  /** Fixed window algorithm: allows to run at most `rate` operations in consecutively segments of duration `per`. */
```
adamw, line 12: that comment is incomplete - please self-review comments

adamw:
@Kamil-Lontkowski thanks, this looks much better. I left some comments - I think it's especially worthwhile describing how the modified algorithms work (for our future selves :) ), and how they differ from the other mode counterparts.

core/src/main/scala/ox/resilience/RateLimiter.scala
uini223: We discussed it with @Kamil, and we came to conclusion that it doesn't make much sense. 
Effectively it would result in "max number of operations currently running"

adamw:
And yes, I think the token bucket only makes sense in the original mode

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
uini223: I would also add info that we did not implement LeakyBucket on purpose

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
uini223: > also, this should be in a finally
Also in Sliding window

core/src/main/scala/ox/resilience/RateLimiter.scala
uini223: So I see that we have 3 options here:
1. Leave it as it is (with Enum as additional parameter)
2. Add new methods in this class e.g. durationFixedWindow
3. Add DurationRateLimiter just to have factory methods `fixedWindow` and `durationWindow` 

For me number 3 should do the job, but that's a bit matter of style. 
@adamw Wdyt?

uini223 (dismissed):

core/src/main/scala/ox/resilience/RateLimiter.scala
adamw: I think I'd keep `RateLimiter` as the single entry point. But separate methods are probably better, as it's an entirely different operation. So let's go with 2.

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
Kamil-Lontkowski: Added

core/src/main/scala/ox/resilience/DurationRateLimiterAlgorithm.scala
Kamil-Lontkowski: Added

core/src/main/scala/ox/resilience/RateLimiter.scala
Kamil-Lontkowski: I added new methods

adamw:
Thanks! :)

PR #257, opened 2024-12-19
Adaptive retries

### LINKED ISSUE #252: Adaptive retries
See: https://www.youtube.com/watch?v=rvHd4Y76-fs & https://github.com/aws/aws-sdk-java-v2/blob/07dbaeb55239a6b7a41e360292d2ea1f84e1451a/core/retries/src/main/java/software/amazon/awssdk/retries/AdaptiveRetryStrategy.java#L4

Kamil-Lontkowski:
Closes #252

Implementation of adaptive retries based on implementation from [aws-sdk-java-v2](https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/retries/AdaptiveRetryStrategy.html). It is backed by simple token bucket, where for every retry we take tokens, and replenish them for every successful operation. This way if we are making for example request for very resource-constrained set of resources, after system failure we don't generate more load by retrying every operation that failed.


Kamil-Lontkowski:
I have a few questions regarding this draft.

1. AdaptiveRetryConfig added few parameters, I am wondering if we want to go in similar direction as before, providing factory methods for simple cases, or make them with all parameters with sensible defaults.
2. Currently when there are no tokens for retry we just fail as if we used all retries. I wonder if we can introduce "blocking mode" so client waits for enough tokens in bucket.
3. There is already implementation of leakyBucket in RateLimiter. I wonder if maybe it would be worth sharing some logic between them, but at the same time it is very simple logic.

lukaszlenart:
What about adding token on failure/giving up?

adamw:
Isn't the "adaptiveness" or "token bucket" an orthogonal concern to the other pieces of configuration that we have currently: `schedule`, `resultPolicy` and `onRetry`? That is - for any values of the schedule etc., we might add (or not) a token bucket, which needs to be consulted before retries. In that scenario, we wouldn't need separate `Config` implementations, just extending the current one?

As for your questions:
1. if we have a single config class, then we should also have factory methods with sensible defaults
2. I think that would be making the retries kind of a rate limiter - and it's a different mechanism. So let's keep those separate.
3. if you could extract a `Bucket` abstraction, great, but not unless straighforward

Kamil-Lontkowski:
> What about adding token on failure/giving up?

I think that it would contradict what TokenBucketRetryCondition and AdaptiveRetryStrategy do. If we add token on failure we would effectively take token for retry and add it back after it fails again. I don't know if I understood that correctly.

adamw:
Thanks for the changes, but I've got yet another variant to consider. Up until now, retries where stateless - it was just a config saying how many time to retry, regardless of other invocations. With the adaptive variant, retries become stateful - the token bucket must be shared. State is usually hard and we have to be cautious not to introduce "surprising" APIs.

That's why maybe it would be better to make the state possible explicit. That is - keep `retry` stateless as it currently is. But, introduce an `AdaptiveRetry(TokenBucket)` class, which would have an `apply` method, mirroring the current `retry`. That way, the stateful instance would be made explicit, and that's what users would have to share. Usage would then be:

```scala
// this will be defined at the top-level, shared among request handlers etc.
val adaptiveRetry = AdaptiveRetry(TokenBucket(...))

// usage:
adaptiveRetry(RetryConfig.backoff(maxRetries, initialDelay, maxDelay))(...)
```

As for the implementation, we might use the callbacks in the retry policy (respecting the callbacks that already are there). Also, shouldn't we specify both `failureCost` and `successReward` for the token bucket?

What do you think? :)

Kamil-Lontkowski:
This WIP is mostly based on mechanism implemented in aws-sdk. There is only exceptionCost. But we can add also successReward. Maybe with new retry `AdaptiveRetry` those can be functions on errors and result allowing even more flexibility(in aws-sdk throttling exception were treated differently).

adamw:
Hm, I think that's a good feature of the AWS SDK - if the call didn't even make it over the network, then we shouldn't count this as an "adaptive failure". So just as we have "isSuccess" etc in the policy, here we could also have one piece of config for adaptive retries - whether the exception should incur the retry penalty

Kamil-Lontkowski:
I am trying two approaches for writing AdaptiveRetry class. One defines methods like `failureCost` as class type params. This approach makes the apply methods take less parameters but it forces us to define E and T params when defining `AdaptiveRetry`. So this seems like a bad idea if we want to reuse policy for different actions. It also does't allow for `retry` alternative
```scala
def apply(config: RetryConfig[E, T])(operation: => T): T
```
because we can't narrow T to `Throwable`.

Second approach pushes definition of this method to apply. This also have some drawbacks. We can't have multiple apply methods with the same defaults, so we need to force user to define these, or hardcode defaults. Two argument lists also makes it hard for compiler to pick proper overlaoded method. In the version I pushed I needed to specify with config E and T explicitly because without it compiler would pick method that returns `T` even though operation returns Either. Also the infered type of result was `Any`(but I don't know if it is from IntelliJ or compiler)

adamw:
But isn't failure / success cost constant? Having them variable seems like a very advanced scenario

Kamil-Lontkowski:
These can be easily changed. I was wondering more if we are okey with signature like this:
```scala
def apply[E, T, F[_]](
      config: RetryConfig[E, T],
      failureCost: Int,
      successReward: Int,
      isFailure: E => Boolean,
      errorMode: ErrorMode[E, F]
  )(operation: => F[T]): F[T]
```
Because in this scenario overloading apply methods is not good idea because things like I mentioned in the comment can sneak by and blow out in runtime. But writing few methods, like `retryEither` should be enough. If so I will write tests and complete docs.

Kamil-Lontkowski:
I think that would be nice middle ground for how much configuration should be provided at every invocation. Maybe running operations on `AdaptiveRetry` instance should have the same api that `retry.scala` provides to avoid confusion with this `apply` method. But that is just matter of method name.

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
```diff
@@ -0,0 +1,119 @@
+package ox.resilience
+
+import ox.{EitherMode, ErrorMode}
+import ox.scheduling.scheduledWithErrorMode
+
+import scala.util.Try
+
+/** Provides mechanism of "adaptive" retries. For every retry we take [[failureCost]] from token bucket and for every success we add back to
+  * the bucket [[successReward]] tokens. One instance can be "shared" across multiple operations against constrained resource. This allows
+  * to retry in case of transient failures and at the same time doesn't produce more load on systemic failure of a resource.
```
adamw, line 10: Some more things worth mentioning in the docs:
* inspiration (Amazon SDK source code + re:Invent talk)
* the fact that this instance is thread-safe

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
```diff
@@ -0,0 +1,119 @@
...
+  *   with different [[failureCost]] for example.
+  * @param failureCost
+  *   Number of tokens to take from [[tokenBucket]] when retrying.
+  * @param successReward
+  *   Number of tokens to add back to [[tokenBucket]] after successful operation.
+  */
+case class AdaptiveRetry(
+    tokenBucket: TokenBucket,
+    failureCost: Int = 1,
+    successReward: Int = 1
```
adamw, line 23: I think the defaults should be different? Sth like 100 for failure, 1 for success? What are the values in the Amazon SDK?

Also, can we introduce a `AdaptiveRetry.default` factory method? That would create an instance with a "default" token bucket capacity

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
```diff
@@ -0,0 +1,119 @@
...
+      isFailure: E => Boolean = (_: E) => true,
+      errorMode: ErrorMode[E, F]
+  )(operation: => F[T]): F[T] =
+    val isWorthRetrying: E => Boolean = (error: E) =>
+      // if we cannot acquire token we short circuit and stop retrying
+      val isWorth = config.resultPolicy.isWorthRetrying(error)
+      if isWorth && isFailure(error) then tokenBucket.tryAcquire(failureCost)
+      else isWorth
+
+    val isSuccess: T => Boolean = (result: T) =>
```
adamw, line 62: Doing side effects in a predicate (`isSuccess`) is a code smell. I'd rather assume that this can be freely called multiple times. Can we maybe simply call the token-release code after `operation` completes successfully?

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
```diff
@@ -0,0 +1,119 @@
...
+    *   - the error `E` in context `F` as returned by the last attempt if the config decides to stop.
+    * @see
+    *   [[scheduledWithErrorMode]]
+    */
+  def apply[E, T, F[_]](
+      config: RetryConfig[E, T],
+      isFailure: E => Boolean = (_: E) => true,
+      errorMode: ErrorMode[E, F]
+  )(operation: => F[T]): F[T] =
+    val isWorthRetrying: E => Boolean = (error: E) =>
```
adamw, line 56: same here - maybe we could use another callback?

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
```diff
@@ -0,0 +1,119 @@
...
+    * @return
+    *   Either:
+    *   - the result of the function if it eventually succeeds, in the context of `F`, as dictated by the error mode.
+    *   - the error `E` in context `F` as returned by the last attempt if the config decides to stop.
+    * @see
+    *   [[scheduledWithErrorMode]]
+    */
+  def apply[E, T, F[_]](
+      config: RetryConfig[E, T],
+      isFailure: E => Boolean = (_: E) => true,
```
adamw, line 53: I think if we get an error value (`E`), that's always a failure. The additional flexibility that we provide in retries, is that success values `T` might also cause a retry - so there's no exception thrown, for example (if E == Throwable), but we get back a value which signals a retry anyway. But if we get an E, that's a failure for sure

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
adamw: maybe ... we should change `onRetry` to `afterAttempt: Attempt => ShouldContinue`

where `case class Attempt(result: Either[E, T], willContinue: Boolean, attemptNumber: Int)` and `enum ShouldContinue: case Yes, No`?

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
Kamil-Lontkowski: In the talk the proportions were something like this. But in the sdk code defaults are bucket size = 500 and cost 5. From what I gathered I think that in sdk success and failure have the same cost.

adamw: Can you share a link to the source code maybe?

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
adamw: another approach: we currently translate retries -> schedule, and hence we get a chain adaptive retries -> retries -> schedule. Maybe we should translate adaptive retries directly to a schedule? And add any missing callbacks there

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
Kamil-Lontkowski: [Default size and cost](https://github.com/aws/aws-sdk-java-v2/blob/07dbaeb55239a6b7a41e360292d2ea1f84e1451a/core/retries/src/main/java/software/amazon/awssdk/retries/DefaultRetryStrategy.java#L93-L94) and [number of token released](https://github.com/aws/aws-sdk-java-v2/blob/07dbaeb55239a6b7a41e360292d2ea1f84e1451a/core/retries/src/main/java/software/amazon/awssdk/retries/internal/BaseRetryStrategy.java#L230)

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
Kamil-Lontkowski: Changed signature to `isFailure: T => Boolean`

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
Kamil-Lontkowski: I added two callbacks to SheduledConfig. `afterSuccess: T => Unit` and `shouldAttempt: Either[E, T] => Boolean`

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
Kamil-Lontkowski: I pushed changes according to this comment. `shouldPayPenaltyCost: T => Boolean = (_: T) => true` decides if penalty will be paid for result T that was decided to be retried.

But I am not sure about the use case. What if we use library that has client side throttling and operation ends with error. It was throttling so we didn't make the call to the service so we should not pay for this retry but now we can't handle this case. Maybe it should be `Either[E, T] => Boolean`. Just a thought, I don't know which case might be more useful.

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
```diff
@@ -0,0 +1,132 @@
...
+    *   - the error `E` in context `F` as returned by the last attempt if the config decides to stop.
+    * @see
+    *   [[scheduledWithErrorMode]]
+    */
+  def retryWithErrorMode[E, T, F[_]](
+      config: RetryConfig[E, T],
+      shouldPayPenaltyCost: T => Boolean = (_: T) => true,
+      errorMode: ErrorMode[E, F]
+  )(operation: => F[T]): F[T] =
+    val shouldAttempt: Either[E, T] => Boolean = result =>
```
adamw, line 59: this implementation comes close, but we're still doing a side-effect in a predicate here
I think we'd need only one new callback, `afterAttempt`, parametrised with the result (E/T), attempt number, and a flag if the operation will be run again. There, we can either pay the reward or the penalty

Kamil-Lontkowski: But we still need somehow to check before attempt if we have enough tokens. So at some point in loop in schedule we need to run `tokenBucket.tryAcquire(failureCost)` to determine if we can retry. I don't know if I understand correctly, but if we only consider bucket in `afterAttempt` and get rid of side effects in predicate we will be adding and removing tokens, but never consider if we have enough to retry.

adamw: Hm yeah I see the problem. I guess it's better to aquire the "retry tokens" immediately after an attempt, and fail early, instead of aquiring them before an attempt (after sleeping). Some tokens might be return meanwhile, but that's an edge case it itself, and if the tokens are already scarce, I guess it won't matter much if we aquire the retry tokens immediately after a failed attempt. Is that also what the Amazon SDK does?

As for the callbacks - maybe we can have a single one, `afterAttempt: Either[E, T] => ScheduleContinue’`? Instead of all the `shouldContinueOnError` etc. The name then is not "callback-like", so it should be clear that it shouldn't be called multiple times (as a predicate can). The return value (a simple Yes/No enum) specifies the intent of the return value. WDYT?

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
```diff
@@ -0,0 +1,132 @@
...
+    * @see
+    *   [[scheduled]]
+    */
+  def retry[T](config: RetryConfig[Throwable, T], shouldPayPenaltyCost: T => Boolean = (_: T) => false)(operation: => T): T =
+    retryWithErrorMode(config, shouldPayPenaltyCost, EitherMode[Throwable])(Try(operation).toEither).fold(throw _, identity)
+
+end AdaptiveRetry
+
+object AdaptiveRetry:
+  def default: AdaptiveRetry = AdaptiveRetry(TokenBucket(500), 5, 1)
```
adamw, line 132: do these follows AWS SDK's defaults?

Kamil-Lontkowski: Size and retry cost yes([Default size and cost](https://github.com/aws/aws-sdk-java-v2/blob/07dbaeb55239a6b7a41e360292d2ea1f84e1451a/core/retries/src/main/java/software/amazon/awssdk/retries/DefaultRetryStrategy.java#L93-L94) and [number of token released](https://github.com/aws/aws-sdk-java-v2/blob/07dbaeb55239a6b7a41e360292d2ea1f84e1451a/core/retries/src/main/java/software/amazon/awssdk/retries/internal/BaseRetryStrategy.java#L230)) but in sdk on success acquired tokens are added back, there isn't different value for successReward contrary to what is said in the talk. One token for success is from previous comment.

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
Kamil-Lontkowski: Amazon SDK indeed first tries to acquire token then wait.

As for the callback I think this would simplify config. Might need to change some things and tests but I think this should work

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
adamw: `but in sdk on success acquired tokens are added back` - we also add tokens on success? I think I don't understand :)

Kamil-Lontkowski: Yes, I mean number of tokens. In sdk we acquire `x` tokens for retry and on we success add back to bucket `x` tokens. We have different value for success and failure, in aws-sdk they are the same. So I meant that if we were to follow aws-sdk defaults we would have `AdaptiveRetry(TokenBucket(500), 5, 5)`. But in the talk and in previous comments it was mentioned that on success we should add less tokens to bucket then we require on retry.

adamw: Ah ok, thanks. Let's keep it this way, we'll see how it works

adamw:
The code examples in the docs didn't have `mdoc:compile-only` and contained multiple errors (I fixed those)

core/src/main/scala/ox/resilience/AdaptiveRetry.scala
adamw: Yeah you're right, `Either[E, T] => Boolean` would be much more useful. Also let's rename this to `shouldPayFailureCost` to keep the vocabulary consistent

PR #262, opened 2024-12-31
Fix `runToFile` not creating a file

### LINKED ISSUE #261: Flows - runToFile does not create a file
Flows method `runToFile` https://github.com/softwaremill/ox/blob/master/core/src/main/scala/ox/flow/FlowIOOps.scala#L65 does not create file when it does not exist. 

Fix the issue

emil-bar:
Closes #261


adamw:
Thanks!

PR #263, opened 2024-12-31
Schedule based on cron expressions

### LINKED ISSUE #249: Create a repeat config basing on a cron expression
Allowing running background crons using our repeat mechanism

Kamil-Lontkowski:
Closes #249

Defines additional `Schedule` to run based on cron expressions defined with [cron4s](https://github.com/alonsodomin/cron4s).


Kamil-Lontkowski:
Regarding this draft I implemented `repeat` methods based on `scheduling.repeat` but for example I don't know if `shouldContinueOnResult` makes sense in this case.

There is also small correction in adaptive retries docs.

cron/src/main/scala/ox/scheduling/cron/cron.scala
```diff
@@ -0,0 +1,72 @@
...
+  * @param operation
+  *   The operation to repeat.
+  * @return
+  *   The result of the last invocation if the config decides to stop.
+  * @throws anything
+  *   The exception thrown by the last invocation if the config decides to stop.
+  * @see
+  *   [[scheduled]]
+  */
+def repeat[T](cron: CronExpr, shouldContinueOnResult: T => Boolean = (_: T) => true)(operation: => T): T =
```
adamw, line 27: can we maybe design the integration so that it's simply yet another `Schedule` which might be plugged into a `RepeatConfig`? That way, we wouldn't have to define variants of the top-level `repeat` etc. methods

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -3,18 +3,18 @@ package ox.scheduling
...
 
-sealed trait Schedule:
+trait Schedule:
   def nextDuration(invocation: Int, lastDuration: Option[FiniteDuration]): FiniteDuration
   def initialDelay: FiniteDuration = Duration.Zero
 
 object Schedule:
 
-  private[scheduling] sealed trait Finite extends Schedule:
+  private[scheduling] trait Finite extends Schedule:
```
adamw, line 12: it would be nice to keep those sealed (I suspect that are some non-exhaustive warnings due to this change). Maybe we can introduce a new infinite-type schedule, which would capture a "compute next invocation" function, and then use this from the cron integration?

This can even be stateful - so that schedules can depend on past history of executions

Kamil-Lontkowski: Then only pattern matching against `Schedule` is done in this code:
```scala
val remainingInvocations = config.schedule match
    case finiteSchedule: Schedule.Finite => Some(finiteSchedule.maxRepeats)
    case _                               => None
```
so there are no warningg, but I kept those sealed.

core/src/main/scala/ox/scheduling/Schedule.scala
```diff
@@ -16,6 +16,13 @@ object Schedule:
 
   private[scheduling] sealed trait Infinite extends Schedule
 
+  private[scheduling] final case class ExternalInfinite(
```
adamw, line 19: maybe let's call it `Computed`? Also, the `computeNextDuration` could return an option, where `None` means we should stop the schedule. And a question - when the parameter to `computeNextDuration` might be `None`? A comment here would be helpful

adamw: Ah to compute the initial duration - nevermind the last point then :)

cron/src/main/scala/ox/scheduling/cron/CronSchedule.scala
```diff
@@ -0,0 +1,40 @@
...
+import cron4s.lib.javatime.*
+import cron4s.{Cron, CronExpr, toDateTimeCronOps}
+import ox.scheduling.Schedule
+
+import java.time.LocalDateTime
+import java.time.temporal.ChronoUnit
+import java.util.concurrent.TimeUnit
+import scala.concurrent.duration.{Duration, FiniteDuration}
+
+/** Methods in this object provide
```
adamw, line 12: provide ...? ;)

core/src/main/scala/ox/scheduling/Schedule.scala
Kamil-Lontkowski: I am not sure about returning `Option`. This only defines schedule, we have other options in configs to stop the schedule and `scheduledWithErrorMode` provides all the options. We would need to change all schedules or treat `Computed` in schedule loop differently. We already have `afterAttempt` in `ScheduledConfig` and I don't think that another way to do the same thing would be clear for the user. 

In `CronSchedule` we have an option because of the `cron.next(now)`, but it should return None only in case like: we want to get next time but supplied `from` is LocalDate, so we don't have info about current time. LocalDateTime have nanosecond precision so we should never get None.

adamw: Ok, let's go with `ComputedInifite` then :)

PR #266, opened 2025-01-09
Circuit breaker

Kamil-Lontkowski:
This draft implements CricuitBreaker with features based on that are provided in breaker from [resilience4j](https://resilience4j.readme.io/docs/circuitbreaker).

Those features, for count based window (last n operations):
- defining threshold for failures
- defining threshold for slow operations, and threshold for what is considered slow
- minimum number of operations until thresholds are calculated
- HalfOpen state that allows number of operations to pass, after that it is calculated if those operations where below threshold so we can close breaker or go back to open.
- time after which breaker goes from Open to HalfOpen

Things that are not implemented in here but are in resilience4j.
- Timeout for HalfOpen state, after which breaker goes back to Open
- Ability to turn of automatic transition from Open to HalfOpen state


core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,186 @@
...
+      val failuresRate = ((results.count(_ == CircuitBreakerResult.Failure) / windowSize.toFloat) * 100).toInt
+      val slowRate = ((results.count(_ == CircuitBreakerResult.Slow) / windowSize.toFloat) * 100).toInt
+      state.updateAndGet { current =>
+        current match
+          // After operation check if we didn't cross thresholds
+          case CircuitBreakerState.Closed =>
+            if numOfOperations >= minimumNumberOfCalls && (failuresRate >= failureRateThreshold || slowRate >= slowCallThreshold) then
+              // Start schedule to switch to HalfOpen after waitDurationOpenState passed
+              forkDiscard(
+                scheduled(ScheduledConfig[Throwable, Unit](Schedule.InitialDelay(waitDurationOpenState)))(
```
Kamil-Lontkowski, line 142: Effects in updateAndGet should be avoided, but I think scheduling state change twice (in case of reapplying function) should not break things, especially since we should have braker open for some time. But maybe it can create race condition where we can complete enough calls in HalfOpen state to change it to Closed only for second scheduled to complete and change it back to HalfOpen

adamw: Hm this looks fragile. I think we should change the design a little. An actor here seems like a natural choice - the actor state - that is, the state machine - would then always be accessed from a single thread. We could send updates to the actor with the results of operation invocations, and basing on the, the internal state would be updated.

One place where we'd have to deviate from the actor pattern is checking if the circuit breaker is open - always going through the actor would be a bottleneck (as all operations would have to synchronize on a single thread). So the actor would also have to update some shared mutable state, which every thread could quickly check (not necessarily immediately consistent, it's fine if we let through one or two additional operations, when the CB is closing)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,186 @@
...
+          config.slowCallDurationThreshold,
+          duration,
+          config.minimumNumberOfCalls,
+          config.numberOfCallsInHalfOpenState,
+          config.waitDurationOpenState
+        )
+  end apply
+
+  private[resilience] case class CircuitBreakerCountStateMachine(
+      windowSize: Int,
```
adamw, line 99: this should definitely be captured in a "config" case class. Can we reuse the existing one?

core/src/main/scala/ox/resilience/CircuitBreaker.scala
adamw: Then each circuit breaker state could be modelled as an immutable value, so we'd only have one top-level mutable overall state, managed by the actor

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: It is captured in config, in enum SlidingWindow

adamw:
As for the questions:
1. hm I don't know - how do circuit breakers work normally? maybe you could find some articles describing typical designs, and link to them here or in the comments; I think we should aim for whatever is "industry standard"
2. again, that's a question of how circuit breakers work in general. Intuitively, the half open state should transition to open/closed after certain conditions are met, but I'm sure there's plenty of edge-cases to consider
3. yeah, `runOrDrop` is fine
4. I think any use-case for wiping can be served by simply creating a new CB, so let's omit wiping for now

core/src/main/scala/ox/resilience/CircuitBreaker.scala
adamw: yes, but here we are passing all these parameters explicitly for some reason? the parameter list seems way too long :)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: I also would need actorRef to schedule state update within state machine but I can't seem to figure out how to pass it since they depend on each other

Kamil-Lontkowski:
Answer based on what is available in [pekko/akka](https://pekko.apache.org/docs/pekko/1.1/common/circuitbreaker.html), [monix](https://monix.io/docs/current/catnap/circuit-breaker.html), [rezilience (zio)](https://github.com/svroonland/rezilience). [circuit (cats-effect)](https://davenverse.github.io/circuit/) and [polly(C#)](https://www.pollydocs.org/strategies/circuit-breaker.html)

1. [resilience4j](https://resilience4j.readme.io/docs/circuitbreaker) provide some more configuration in this regard than other libs. I think it would be best to calculate metrics for different states separately. Other libs allow just for one call and if it succeeds it closes.
2. This also seems like more flexibility on resilience4j side.

Monix, circuit and Pekko/Akka works exactly the same. They count failures (or slow calls) in a row not a rate based on window. Then wait before going to halfOpen and then deciding based on one operation result. Plus the wait duration before transitioning to halfOpen is configured as backoff.

rezilience provides maxFailures in a row just like monix and also count based sliding window. It also supports different schedules for waiting before going to halfOpen state. It also allows only one call to decide if it goes back to open or close.

Polly is little different but only in few cases. It provides threshold rates for sampling window of some duration. As I understand it in effect means sliding window(But maybe it is simpler and works just like fixed window). It also support minimum number of calls to be able trip in a sample. It also allows for dynamically determining break duration before switching to halfOpen.It also provide ability to set state manually and reading current state through CircuitBreakerStateProvider.

Zio, resilience4j and rezilience also provides ability to consume different events like state changes or metrics.

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,247 @@
...
+  case Closed
+  case HalfOpen(since: Long, semaphore: Semaphore, completedOperations: Int = 0)
+
+enum CircuitBreakerResult:
+  case Success
+  case Failure
+  case Slow
+
+case class Metrics(
+    failureRate: Int,
```
adamw, line 24: it would be good to name it or at least comment to be more precise on what "rate" is - I assume calls per second? or another unit? :)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,247 @@
...
+        Some(result)
+      else
+        actorRef.tell(_.registerResult(CircuitBreakerResult.Failure, acquiredResult))
+        Some(result)
+      end if
+    else None
+    end if
+  end runOrDrop
+
+  def runEitherOrDrop[E, T](resultPolicy: ResultPolicy[E, T] = ResultPolicy.default[E, T])(
```
adamw, line 100: shouldn't this just delegate to `runOrDrop` using the either error mode?

Also we need the "basic" case of `runOrDrop(op: => T): T` where failures are exceptions

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,247 @@
...
+    case CircuitBreakerState.Closed                                => AcquireResult(true, CircuitBreakerState.Closed)
+    case currState @ CircuitBreakerState.Open(_)                   => AcquireResult(false, currState)
+    case currState @ CircuitBreakerState.HalfOpen(_, semaphore, _) => AcquireResult(semaphore.tryAcquire(1), currState)
+
+  def runOrDrop[E, F[_], T](em: ErrorMode[E, F], resultPolicy: ResultPolicy[E, T] = ResultPolicy.default[E, T])(op: => F[T]): Option[F[T]] =
+    val acquiredResult = tryAcquire
+    if acquiredResult.acquired then
+      val before = System.nanoTime()
+      val result = op
+      val after = System.nanoTime()
```
adamw, line 82: we should have a `timed` utility method in ox, I think we have such time measurements in a couple of places

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,247 @@
...
+    )
+    assert(
+      config.slowCallThreshold >= 0 && config.slowCallThreshold <= 100,
+      s"slowCallThreshold must be between 0 and 100, value: ${config.slowCallThreshold}"
+    )
+
+    private val callResults: Array[Option[CircuitBreakerResult]] = Array.fill[Option[CircuitBreakerResult]](windowSize)(None)
+    private var writeIndex = 0
+
+    private var _state: CircuitBreakerState = CircuitBreakerState.Closed
```
adamw, line 168: this is accessed externally, so needs to be concurrency-protected? with a large disclaimer that although this does break the actor's internal state protection, we know what we're doing ;)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,247 @@
...
+
+    def state: CircuitBreakerState = _state
+
+    def registerResult(result: CircuitBreakerResult, acquired: AcquireResult): Unit =
+      callResults(writeIndex) = Some(result)
```
adamw, line 173: in fact, it might be good to separate the "pure" state machine update function - which takes the current state + call result as a parameter, and outputs the new state + optional self-callbacks to run (?); from the mutable actor's state, which can be queried externally.

adamw:
> Answer based on what is available in [pekko/akka](https://pekko.apache.org/docs/pekko/1.1/common/circuitbreaker.html), [monix](https://monix.io/docs/current/catnap/circuit-breaker.html), [rezilience (zio)](https://github.com/svroonland/rezilience). [circuit (cats-effect)](https://davenverse.github.io/circuit/) and [polly(C#)](https://www.pollydocs.org/strategies/circuit-breaker.html)

Good analysis, thanks :) So basing on that, what design would you propose? What would be the configuration options, and the algorithm of transferring between closed/ho/open states?

Not sure if we need both count-based and windowed variants - isn't the count-based variant a windowed variant, but with window duration = Inf?

core/src/test/scala/ox/resilience/CircuitBreakerTest.scala
```diff
@@ -0,0 +1,59 @@
+package ox.resilience
+
+import org.scalatest.flatspec.AnyFlatSpec
+import org.scalatest.matchers.should.Matchers
+import scala.concurrent.duration.*
+import ox.*
+
+class CircuitBreakerTest extends AnyFlatSpec with Matchers:
```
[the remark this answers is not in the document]
adamw: Ah yes, this might be a good idea - but I would then double-down on separating the "pure" and side-effecting/mutating parts. And write a lot of tests for the pure part, and some "integration" ones for the whole package

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: I don't really know why I didn't do that, will fix.

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: Honestly I don't think we have. We have couple of similar methods as test utilities.

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: It's percent of failed call in last window. I wanted to move state logic to parent trait and only leave calculating metrics to different implementations

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: I was wondering if it needs to be atomic. We read concurrently but write only from the same actor's thread

adamw: Concurrent reads need to be protected as well. If you don't know why, it might be good to read on memory barriers, java's memory model and concurrency primitives :)

adamw:
But in a count-based approach, you're counting all results anyway?

Kamil-Lontkowski:
I changed calculating state change to be pure function. Added docs in code and some tests for the state machine. I will add more tests especially testing if schedules and timeout for state changes work properly. I also need to add docs, I think in this case some kind of diagram would be helpful to understand how this work.

Kamil-Lontkowski:
Also seems like CI is broken because of deprecation of `actions/upload-artifact: v3`

adamw:
> Also seems like CI is broken because of deprecation of actions/upload-artifact: v3

Let's fix it then, in a separate PR then :)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,121 @@
...
+      waitDurationOpenState = c.waitDurationOpenState,
+      halfOpenTimeoutDuration = c.halfOpenTimeoutDuration
+    )
+end CircuitBreakerStateMachineConfig
+
+/** Circuit Breaker. Operations can be dropped, when the breaker is open or if it doesn't take more operation in halfOpen state. The Circuit
+  * Breaker might calculate different metrics based on [[SlidingWindow]] provided in config. See [[SlidingWindow]] for more details.
+  */
+class CircuitBreaker(val config: CircuitBreakerConfig)(using Ox):
+  val stateMachine = CircuitBreakerStateMachine(config)
```
adamw, line 57: should be private?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,192 @@
...
+
+import scala.concurrent.duration.*
+import ox.*
+import ox.scheduling.scheduled
+import ox.scheduling.{ScheduledConfig, Schedule}
+import java.util.concurrent.Semaphore
+import ox.channels.ActorRef
+
+private sealed trait CircuitBreakerStateMachine(val config: CircuitBreakerStateMachineConfig)(using val ox: Ox):
+  def calculateMetrics(lastAcquisitionResult: Option[AcquireResult], timestamp: Long): Metrics
```
adamw, line 11: do we need to expose these methods as public? ideally, the interface exposed for the actor would be only the ones that are used via actorRef

core/src/main/scala/ox/resilience/AtomicCircularBuffer.scala
```diff
@@ -0,0 +1,29 @@
+package ox.resilience
+
+import scala.reflect.ClassTag
+import java.util.concurrent.Semaphore
+
+class AtomicCircularBuffer[T: ClassTag](size: Int):
```
adamw, line 6: this is not used?

core/src/test/scala/ox/resilience/CircuitBreakerTest.scala
```diff
@@ -0,0 +1,36 @@
...
+  behavior of "Circuit Breaker"
+
+  it should "drop operation after exceeding threshold" in supervised {
```
adamw, line 11: a positive test-case would be good to have as well

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,192 @@
...
+
+    private val callResults: Array[Option[CircuitBreakerResult]] = Array.fill[Option[CircuitBreakerResult]](windowSize)(None)
+    private var writeIndex = 0
+
+    def onStateChange(oldState: CircuitBreakerState, newState: CircuitBreakerState): Unit =
+      import CircuitBreakerState.*
+      // we have to match so we don't reset result when for example incrementing completed calls in halfopen state
+      (oldState, newState) match
+        case (Closed, Open(_) | HalfOpen(_, _, _)) =>
+          callResults.mapInPlace(_ => None).discard
```
adamw, line 122: is this an "array wipe"? could be factored out to a method, and written using a simple loop, not mapping

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,192 @@
...
+  def onStateChange(oldState: CircuitBreakerState, newState: CircuitBreakerState): Unit
+
+  @volatile private var _state: CircuitBreakerState = CircuitBreakerState.Closed
+
+  def state: CircuitBreakerState = _state
+
+  def registerResult(result: CircuitBreakerResult, acquired: AcquireResult, selfRef: ActorRef[CircuitBreakerStateMachine]): Unit =
+    updateResults(result)
+    val oldState = _state
+    val newState = nextState(calculateMetrics(Some(acquired), System.currentTimeMillis()), oldState)
```
adamw, line 22: won't calculating metrics on every call be expensive? or rather: calculating metrics this way, iterating over all results. Especially if the window sizes are larger than 100, say 1000 - that's 1000 ops for every call

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,192 @@
...
+    _state = newState
+    scheduleCallback(oldState, newState, selfRef)
+    onStateChange(oldState, newState)
+  end registerResult
+
+  def updateState(selfRef: ActorRef[CircuitBreakerStateMachine]): Unit =
```
adamw, line 28: shouldn't registerResult simply call this method?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,192 @@
...
+        // schedule switch to halfOpen after timeout
+        updateAfter(config.waitDurationOpenState, selfRef)
+      case (CircuitBreakerState.Open(_), CircuitBreakerState.HalfOpen(since, semaphore, completedOperations)) =>
+        // schedule timeout for halfOpen state if is not 0
+        if config.halfOpenTimeoutDuration.toMillis != 0 then updateAfter(config.halfOpenTimeoutDuration, selfRef)
+      case _ => ()
+
+  private def updateAfter(after: FiniteDuration, actorRef: ActorRef[CircuitBreakerStateMachine])(using Ox): Unit =
+    forkDiscard:
+      scheduled(ScheduledConfig(Schedule.InitialDelay(after)))(actorRef.tell(_.updateState(actorRef)))
```
adamw, line 51: is this just to schedule something with a dealy? if so, sleep + tell would be much simpler

adamw:
Looks close to completion :) We'll definitely need some docs

core/src/main/scala/ox/resilience/CircuitBreaker.scala
adamw: so ... let's add one? ;)

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,192 @@
...
+
+  private[resilience] def nextState(metrics: Metrics, currentState: CircuitBreakerState): CircuitBreakerState =
```
adamw, line 53: maybe this method can go to the `object`? it would be clear then that the method doesn't use any of the internal state

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,192 @@
...
+        failuresRate,
+        slowRate,
+        numOfOperations,
+        lastAcquisitionResult,
+        timestamp
+      )
+    end calculateMetrics
+  end CircuitBreakerCountStateMachine
+
+  private[resilience] case class CircuitBreakerTimeStateMachine(
```
adamw, line 153: maybe instead of inhertiance, we could use composition: factor out this state into a `CBResults` class, which would be the state held by `CBStateMachine`. I think it would be easier to understand then, also minimizing the scope of dependencies

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,247 @@
...
+  case Slow
+
+case class Metrics(
+    failureRate: Int,
+    slowCallsRate: Int,
+    operationsInWindow: Int
+)
+
+enum SlidingWindow:
+  case CountBased(windowSize: Int)
```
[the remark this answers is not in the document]
Kamil-Lontkowski: I don't really know what alias can be more intuitive - maybe WindowCapacity?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
Kamil-Lontkowski: Done

core/src/main/scala/ox/resilience/AtomicCircularBuffer.scala
Kamil-Lontkowski: I though I already deleted it

core/src/main/scala/ox/resilience/CircuitBreaker.scala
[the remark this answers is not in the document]
adamw: yeah sorry, here I guess an Int is fine. I think I added the comment in the wrong place, it's the other `Int` occurrences that were more misleading

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
adamw: Hm can't we do a rolling update for time windows as well? We know how much time passed, so which elements to drop (might be 0-*), and which to add?

core/src/main/scala/ox/resilience/CircuitBreaker.scala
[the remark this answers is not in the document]
Kamil-Lontkowski: I added PercentageThreshold type for thresholds, I think minimumNumberOfCalls and numberOfCallsInHalfOpenState are descriptive enough

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
Kamil-Lontkowski: Changed it in last commit. removeHeadWhile short-circuit on first negative and return removed elements so we can inspect them and update stats

adamw:
I added some comments with some questions to consider what the behaviour should be in case of concurrently running operations. I think it would be good to describe the algorithm (in a comment), detailing exactly when state transitions occur, depending on configuration, and what happens if there are other operations who report their results later. We should should be ready for hundreds/thousands of such concurrently running operations reporting failures. So before fixing this (and other) tests, maybe let's consider some possibilities as to how to best fix the problem.

Kamil-Lontkowski:
Regarding the failing test, the assumption is wrong. It should not go back to closed after the halfOpenTimeoutDuration passes. This timeout is guarding slow operations, so if we have halfOpen state and operations are slow we open again if it exceeds timeout. Only way to go from HalfOpen to Closed is to complete `numberOfCallsInHalfOpenState` within timeout duration and without exceeding any threshold.

But I think in case of many concurrent operations we have cases where some operation could break statistics. For example operation started in Closed state that finishes in HalfOpen state will not count towards finished operations in HalfOpen state but will be included in rates. I think we can guard against it by checking if `AcquireResult` has the same state that we are in currently. It will not work for cases where operation last for example whole cycle of changing states. But it can also be done - just add `since` timestamp in Open state and make distinction based on that. So it will just ignore results that were started in different state. All statistics are cleared on change state anyway so we just need to not count the ones started in different state.

adamw:
@Kamil-Lontkowski ok, can you then first fix my test case (fixing the assertions/ checking that it eventually does transit to HO), and add a test case for the scenario you describe? So that we first have a failing test case, and then we can proceed to fixing the problem

adamw:
And even more so, a description of when state transitions occur is crucial :) (for understanding & using the CB)

Kamil-Lontkowski:
For now the part in docs looks like that
image

Maybe these colors are a bit much, but since I had mistaken Closed and Open state many times I wanted to make the Closed pop as the one allowing operations.
This was made with [mermaidjs](https://mermaid.js.org/syntax/stateDiagram.html)
```
stateDiagram
  direction LR
  classDef red fill:red
  classDef green fill:green
  classDef yellow fill:yellow

  Closed --> Open: 1
  Closed --> HalfOpen: 2
  Open --> HalfOpen: 3
  HalfOpen --> Open: 4
  HalfOpen --> Closed: 5

  class Closed green
  class Open red
  class HalfOpen yellow
```

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
+# Circuit Breaker
```
adamw, line 1: there's a few typos in the text below - if you're using VS Code you might try the spell checker plugin, it does quite a good job fixing such problems

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+```
+
+## API
+
+```scala mdoc:compile-only
+import ox.supervised
+import ox.resilience.*
+
+supervised:
+  val circuitBreaker = CircuitBreaker(CircuitBreakerConfig.default)
```
adamw, line 17: do you need to pass the config, even if it's the default one?

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
+# Circuit Breaker
+
+The circuit breaker allows controlling execution of operations and stops if certain condition are met. CircuitBreaker is thread-safe and uses [actor](./actors.md) underneath to change breaker state.
```
adamw, line 3: I think a slightly longer introduction to what a CB does would be in order. Or a link to some external resource explaining what it is?

The actor is a technical detail which can be mentioned in some further section.

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+
+There are two ways that metrics are calculated.
+
+- Count based sliding window - `SlidingWindow.CountBased`, counts metrics based on last n call results.
+- Time based sliding window - `SlidingWindow.TimeBased`, counts metrics based on call results recorded in the lapse of duration before current time.
+
+### Failure rate and slow call rate thresholds
+
+The state of the CircuitBreaker changes from `Closed` to `Open` when the `failureRate` is greater or equal to configurable threshold. For example when 80% of recorded call results failed.
+Failures are counted based on provided `ErrorMode`.
```
adamw, line 37: (...) E.g. any exception that is thrown by the operation, when using the direct, "unwrapped" API.

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+
+The same state change also happen when percentage of slow calls (exceeding `slowCallDurationThreshold`) is equal or greater than configured threshold. For examaple 80% of calls took longer then 10 seconds.
```
adamw, line 39: here I suspect what is considered "slow" is also configurable?

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+- `waitDurationOpenState: FiniteDuration` - duration that CircuitBreaker will wait before switching from `Open` state to `HalfOpen`.
+- `halfOpenTimeoutDuration: FiniteDuration` - timeout for `HalfOpen` state after which, if not enough calls were recorder, breaker will go back to `Open` state. Zero means there is no timeout.
+- `numberOfCallsInHalfOpenState: Int` - number of calls recorded in `HalfOpen` state needed to calculate metrics to decide if breaker should go back to `Open` state or `Closed`. It is also maximum number of operations that can be started in this state.
+
+`SlidingWindow` variants:
+
+- `CountBased(windowSize: Int)` - This variant calculates metrics based on last n results of calls recorded. These statistics are cleared on every state change.
+- `TimeBased(duration: FiniteDuration)` - This variant calculates metrics of operations in the lapse of `duraiton` before current time. These statistics are cleared on every state change.
+
+Values defined in `CircuitBreakerConfig.default`:
```
adamw, line 59: Let's make this a separate section, also explaining how to provide custom config (with a code example)

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+The state of the CircuitBreaker changes from `Closed` to `Open` when the `failureRate` is greater or equal to configurable threshold. For example when 80% of recorded call results failed.
+Failures are counted based on provided `ErrorMode`.
+
+The same state change also happen when percentage of slow calls (exceeding `slowCallDurationThreshold`) is equal or greater than configured threshold. For examaple 80% of calls took longer then 10 seconds.
+
+Those metrics are considered only when number of recorder calls is greater or equal to `minimumNumberOfCalls`, otherwise we don't change state even if `failureRate` is 100%.
+
+### Parameters
+
+- `failureRateThreshold: PercentageThreshold` - percentage of recorder calls marked as failed required to switch to open state.
```
adamw, line 45: [typo] recorded, not recorder?

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+
+- `CountBased(windowSize: Int)` - This variant calculates metrics based on last n results of calls recorded. These statistics are cleared on every state change.
+- `TimeBased(duration: FiniteDuration)` - This variant calculates metrics of operations in the lapse of `duraiton` before current time. These statistics are cleared on every state change.
+
+Values defined in `CircuitBreakerConfig.default`:
+
+```
+failureRateThreshold = PercentageThreshold(50)
+slowCallThreshold = PercentageThreshold(50)
+slowCallDurationThreshold = 60.seconds
```
adamw, line 64: isn't that a bit high for a default? what are the defaults in other systems?

doc/utils/circuit-breaker.md
```diff
@@ -0,0 +1,110 @@
...
+slidingWindow = SlidingWindow.CountBased(100)
+minimumNumberOfCalls = 20
+waitDurationOpenState = FiniteDuration(10, TimeUnit.SECONDS)
+halfOpenTimeoutDuration = FiniteDuration(0, TimeUnit.MILLISECONDS)
+numberOfCallsInHalfOpenState = 10
+```
+
+## Conditions for state change
+
+![state diagram](/_static/state-diagram-cb.svg)
```
adamw, line 74: good idea with the diagram :)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,118 @@
...
+    slowCallsRate: Int,
+    operationsInWindow: Int,
+    lastAcquisitionResult: Option[AcquireResult],
+    timestamp: Long
+)
+
+private[resilience] case class AcquireResult(acquired: Boolean, circuitState: CircuitBreakerState)
+
+private case class CircuitBreakerStateMachineConfig(
+    failureRateThreshold: Int,
```
adamw, line 32: shouldn't we use PercentageThreshold here as well?

core/src/main/scala/ox/resilience/CircuitBreaker.scala
adamw: bump ;)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,118 @@
...
+      halfOpenTimeoutDuration = c.halfOpenTimeoutDuration
+    )
+end CircuitBreakerStateMachineConfig
+
+/** Circuit Breaker. Operations can be dropped, when the breaker is open or if it doesn't take more operation in halfOpen state. The Circuit
+  * Breaker might calculate different metrics based on [[SlidingWindow]] provided in config. See [[SlidingWindow]] for more details.
+  */
+case class CircuitBreaker(config: CircuitBreakerConfig)(using Ox):
+  private[resilience] val stateMachine = CircuitBreakerStateMachine(config)
+  private val actorRef: ActorRef[CircuitBreakerStateMachine] = Actor.create(stateMachine)(using sc = BufferCapacity.apply(100))
```
adamw, line 58: maybe the CB constructor could also accept a BufferCapacity parameter, to allow configuring how much operations to buffer; we could provide a default of 100 there

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,118 @@
...
+
+  private def tryAcquire: AcquireResult = stateMachine.state match
```
adamw, line 60: by convention, side-effecting parameterless operations should have `()`, so: `tryAcquire()`; but `hasPermits` would have no `()`, as it doesn't change the state

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+end CircuitBreakerStateMachine
+
+private[resilience] sealed trait CircuitBreakerResults(using val ox: Ox):
+  def onStateChange(oldState: CircuitBreakerState, newState: CircuitBreakerState): Unit
+  def updateResults(result: CircuitBreakerResult): Unit
+  def calculateMetrics(lastAcquisitionResult: Option[AcquireResult], timestamp: Long): Metrics
+
+private[resilience] object CircuitBreakerResults:
+  case class CountBased(windowSize: Int)(using ox: Ox) extends CircuitBreakerResults(using ox):
+    private val results = new collection.mutable.ArrayDeque[CircuitBreakerResult](windowSize)
```
adamw, line 127: [minor shouldn't  this have expected size `windowsSize+1`? we're removing elements after there's one more element

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+        resultingQueue.removeHeadOption(false) match
+          case Some(CircuitBreakerResult.Success) => successCalls -= 1
+          case Some(CircuitBreakerResult.Failure) => failedCalls -= 1
+          case Some(CircuitBreakerResult.Slow)    => slowCalls -= 1
+          case None                               => ()
+    end updateResults
+
+    def calculateMetrics(lastAcquisitionResult: Option[AcquireResult], timestamp: Long): Metrics =
+      val numOfOperations = results.length
+      val failuresRate = ((failedCalls / numOfOperations.toFloat) * 100).toInt
```
adamw, line 168: [minor] this could be probably made into a smart constructor e.g. `Percentage.of(a, b)`

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+
+    private def clearResults: Unit =
+      results.clear()
+      slowCalls = 0
+      failedCalls = 0
+      successCalls = 0
+
+    def onStateChange(oldState: CircuitBreakerState, newState: CircuitBreakerState): Unit =
+      import CircuitBreakerState.*
+      // we have to match so we don't reset result when for example incrementing completed calls in halfopen state
```
adamw, line 140: so .... we clear results when the state changes? maybe we could introduce a method on the state e.g. `isSameState(other)` to make this clearer?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+    private val results: CircuitBreakerResults
+)(using val ox: Ox):
+  @volatile private var _state: CircuitBreakerState = CircuitBreakerState.Closed(System.currentTimeMillis())
+
+  def state: CircuitBreakerState = _state
+
+  def registerResult(result: CircuitBreakerResult, acquired: AcquireResult, selfRef: ActorRef[CircuitBreakerStateMachine]): Unit =
+    // We check that result was acquired in the same state that we are currently in
+    val isResultFromCurrentState = (acquired.circuitState, _state) match
+      case (CircuitBreakerState.Open(sinceOpen), CircuitBreakerState.Open(since)) if since == sinceOpen                             => true
```
adamw, line 20: `isSameState` then could be used here as well?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+      case _                                                                          => 0
+    val exceededThreshold = (metrics.failureRate >= config.failureRateThreshold || metrics.slowCallsRate >= config.slowCallThreshold)
+    val minCallsRecorder = metrics.operationsInWindow >= config.minimumNumberOfCalls
+    currentState match
+      case CircuitBreakerState.Closed(since) =>
+        if minCallsRecorder && exceededThreshold then
+          if config.waitDurationOpenState.toMillis == 0 then
+            CircuitBreakerState.HalfOpen(currentTimestamp, Semaphore(config.numberOfCallsInHalfOpenState))
+          else CircuitBreakerState.Open(currentTimestamp)
+        else CircuitBreakerState.Closed(since)
```
adamw, line 91: you could pattern match to `self@Closed(...)` and then return `self` here, to make it clear that there is not state change

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
adamw: but overall, good way of filtering out old results :)

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+      case CircuitBreakerState.Open(since) =>
+        val timePassed = (currentTimestamp - since) >= config.waitDurationOpenState.toMillis
+        if timePassed || config.waitDurationOpenState.toMillis == 0 then
```
adamw, line 94: should we ever be in `Open` if `waitDurationOpenState` == 0? Also, `timePassed` will always be >= 0, so this is a no-op?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+        CircuitBreakerStateMachine(
+          CircuitBreakerStateMachineConfig.fromConfig(config),
+          CircuitBreakerResults.TimeWindowBased(duration)
+        )
+  end apply
+
+  def nextState(metrics: Metrics, currentState: CircuitBreakerState, config: CircuitBreakerStateMachineConfig): CircuitBreakerState =
+    val currentTimestamp = metrics.timestamp
+    // We want to know if last result should be added to completed calls in halfOpen state
+    val lastCompletedCall = metrics.lastAcquisitionResult match
```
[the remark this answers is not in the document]
adamw: but because of the initial filtering, is it possible that this is `0`? wouldn't a result from another state be discarded earlier?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+        if !allCallsInHalfOpenCompleted && config.halfOpenTimeoutDuration.toMillis != 0 && timePassed then
+          CircuitBreakerState.Open(currentTimestamp)
+        // If halfOpen calls were completed && rates are below we close breaker
+        else if allCallsInHalfOpenCompleted && !exceededThreshold then CircuitBreakerState.Closed(currentTimestamp)
+        // If halfOpen calls completed, but rates are still above go back to open
+        else if allCallsInHalfOpenCompleted && exceededThreshold then CircuitBreakerState.Open(currentTimestamp)
+        // We didn't complete all half open calls, keep halfOpen
+        else
+          metrics.lastAcquisitionResult match
+            case Some(AcquireResult(true, CircuitBreakerState.HalfOpen(s, _, _)))
```
adamw, line 110: same here, we already checked this at the beginning?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,235 @@
...
+  def registerResult(result: CircuitBreakerResult, acquired: AcquireResult, selfRef: ActorRef[CircuitBreakerStateMachine]): Unit =
+    // We check that result was acquired in the same state that we are currently in
+    val isResultFromCurrentState = (acquired.circuitState, _state) match
+      case (CircuitBreakerState.Open(sinceOpen), CircuitBreakerState.Open(since)) if since == sinceOpen                             => true
+      case (CircuitBreakerState.HalfOpen(sinceHalfOpen, _, _), CircuitBreakerState.HalfOpen(since, _, _)) if sinceHalfOpen == since => true
+      case (CircuitBreakerState.Closed(sinceClosed), CircuitBreakerState.Closed(since)) if sinceClosed == since                     => true
+      case _                                                                                                                        => false
+    // If acquired in different state we don't update results
+    if isResultFromCurrentState then results.updateResults(result)
+    updateState(selfRef, Some(acquired))
```
adamw, line 26: ah ... shouldn't we only call `updateState` if the result isn't discarded? otherwise it should be a no-op?

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: I really couldn't decide. This is internal config, so I would need to unwrap it everywhere where it's used.

doc/utils/circuit-breaker.md
Kamil-Lontkowski: This was default in resilience4j, maybe more sane default would be 10 or 5 seconds? Others don't have defaults.

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
Kamil-Lontkowski: No, this is redundant. I `waitDurationOpenState` is 0 then this case will never happen, we'll go straight from Closed to HalfOpen

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: Changed to use it here as well, I also added extension to PercentageThreshold: `def isExceeded(by: Int): Boolean`

doc/utils/circuit-breaker.md
Kamil-Lontkowski: Yes, config always must be passed. Maybe this is a little redundant since we provide default, but I am not sure if providing default implicitly is better.

Kamil-Lontkowski:
I think I addressed everything in this round of comments :). Turns out there were also typos in comments and tests, so I corrected them.

doc/utils/circuit-breaker.md
adamw: Hm tough question ;) My gut feeling is that 60 is really long, so maybe 10 will be better.

doc/utils/circuit-breaker.md
adamw: Hm, well it wouldn't be implicit, just default. As in overloaded methods. I think we're doing that for retries and other utilities, so here we could have a no-arg constructor as well

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,205 @@
...
+    config: CircuitBreakerStateMachineConfig,
+    private val results: CircuitBreakerResults
+)(using val ox: Ox):
+  @volatile private var _state: CircuitBreakerState = CircuitBreakerState.Closed(System.currentTimeMillis())
+
+  def state: CircuitBreakerState = _state
+
+  def registerResult(result: CircuitBreakerResult, acquired: AcquireResult, selfRef: ActorRef[CircuitBreakerStateMachine]): Unit =
+    // If acquired in different state we don't update results
+    if acquired.circuitState.isSameState(_state) then
```
adamw, line 19: now it reads much better :)

core/src/main/scala/ox/resilience/CircuitBreaker.scala
```diff
@@ -0,0 +1,128 @@
...
+      case _                                                                                => false
+end CircuitBreakerState
+
+private[resilience] enum CircuitBreakerResult:
+  case Success
+  case Failure
+  case Slow
+
+private[resilience] case class Metrics(
+    failureRate: Int,
```
adamw, line 29: Still, what unit is the rate in? ;) a comment at least would be helpful, or proper name e.g. `failureRatePerSecond` or `failureRate: PerSecond`

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,205 @@
...
+          CircuitBreakerStateMachineConfig.fromConfig(config),
+          CircuitBreakerResults.TimeWindowBased(duration)
+        )
+  end apply
+
+  def nextState(metrics: Metrics, currentState: CircuitBreakerState, config: CircuitBreakerStateMachineConfig): CircuitBreakerState =
+    val currentTimestamp = metrics.timestamp
+    val exceededThreshold =
+      config.failureRateThreshold.isExceeded(metrics.failureRate) || config.slowCallThreshold.isExceeded(metrics.slowCallsRate)
+    val minCallsRecorder = metrics.operationsInWindow >= config.minimumNumberOfCalls
```
adamw, line 76: recroded ?

adamw: that is, `recorded` ;)

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,205 @@
...
+    if acquired.circuitState.isSameState(_state) then
+      results.updateResults(result)
+      updateState(selfRef, Some(acquired))
+  end registerResult
+
+  def updateState(selfRef: ActorRef[CircuitBreakerStateMachine], acquiredResult: Option[AcquireResult] = None): Unit =
+    val oldState = _state
+    val newState = nextState(results.calculateMetrics(acquiredResult, System.currentTimeMillis()), oldState, config)
+    _state = newState
+    scheduleCallback(oldState, newState, selfRef)
```
adamw, line 28: shouldn't we call `scheduleCallback` and `onStateChange` only if `!oldState.isSameState(newState)` ? that is move the conditional from the `result`s over here

core/src/main/scala/ox/resilience/CircuitBreaker.scala
Kamil-Lontkowski: Maybe here it would be better as `failurePercentage`. It does not have to do anything with time so maybe the word rate is bad idea

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
Kamil-Lontkowski: Something like this? I am not sure I understood
```scala
  def registerResult(result: CircuitBreakerResult, acquired: AcquireResult, selfRef: ActorRef[CircuitBreakerStateMachine]): Unit =
    // If acquired in different state we don't update results
    if acquired.circuitState.isSameState(_state) then
      results.updateResults(result)
    updateState(selfRef, Some(acquired))
  end registerResult

  def updateState(selfRef: ActorRef[CircuitBreakerStateMachine], acquiredResult: Option[AcquireResult] = None): Unit =
    val oldState = _state
    val newState = nextState(results.calculateMetrics(acquiredResult, System.currentTimeMillis()), oldState, config)
    _state = newState
    if !oldState.isSameState(newState) then
      scheduleCallback(oldState, newState, selfRef)
      results.onStateChange(oldState, newState)
```
Always call `updateState` and don't call `scheduleCallback` and `results.onStateChange` if state didn't change

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
```diff
@@ -0,0 +1,205 @@
...
+    (oldState, newState) match
+      case (_, CircuitBreakerState.Open(_)) =>
+        // schedule switch to halfOpen after timeout
+        updateAfter(config.waitDurationOpenState, selfRef)
+      case (
+            CircuitBreakerState.Open(_) | CircuitBreakerState.Closed(_),
+            CircuitBreakerState.HalfOpen(_, _, _)
+          ) =>
+        // schedule timeout for halfOpen state if is not 0
+        if config.halfOpenTimeoutDuration.toMillis != 0 then updateAfter(config.halfOpenTimeoutDuration, selfRef)
```
adamw, line 45: if it's 0, than we should never transition to HO? so this could be an `assert`?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
adamw: Just the second change:

```
 if !oldState.isSameState(newState) then
      scheduleCallback(oldState, newState, selfRef)
      results.onStateChange(oldState, newState)
```

Is there any point in calling the callback / onStateChange if the state did not change?

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
adamw: ah ok, so it's a feature, understood :)

adamw: maybe just add a comment with what you wrote above ;) 

```
// schedule timeout for halfOpen state if is not 0 - to leave HalfOpen state all required calls must complete (as success or failure)
```

core/src/main/scala/ox/resilience/CircuitBreakerStateMachine.scala
Kamil-Lontkowski: No

PR #269, opened 2025-01-15
Bulkhead

Kamil-Lontkowski:
Added Bulkhead based on semaphore. 

[resilience4j](https://resilience4j.readme.io/docs/bulkhead) provides simple bulkhead, [rezilience](https://www.vroste.nl/rezilience/docs/bulkhead/) on top of that allows to setup queue that enables for some number of calls wait until other calls complete.

I added runOrDropWithTimeout that will block until there is permit or timeout passes. It uses `Semaphore.tryAcquire` but it's also possible to implement using `timeout`:
```scala
def runOrDropWithTimeout[T](timeoutDuration: FiniteDuration)(operation: => T): Option[T] =
  timeoutOption(timeoutDuration)(semaphore.acquire())
    .map: _ =>
      try operation
      finally semaphore.release()
```

### DIFF
--- a/core/src/main/scala/ox/resilience/Bulkhead.scala
+++ b/core/src/main/scala/ox/resilience/Bulkhead.scala
@@ -1,0 +1,17 @@
+package ox.resilience
+
+import java.util.concurrent.Semaphore
+import scala.concurrent.duration.*
+
+case class Bulkhead(maxConcurrentCalls: Int):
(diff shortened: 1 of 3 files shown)


adamw:
To be honest, since it's simply a semaphore wrapper, I don't think there's much sense in adding such an abstraction to Ox. People could just use a `Semaphore` with the same effect.

PR #317, opened 2025-05-29
Replace the Schedule used for repeats/retries with an interval-list based approach

adamw:
The  new `Schedule` class allows for compositional definitions of schedules, and follows the design outlined by @miciek  in https://www.youtube.com/watch?v=RWvTziDfFUQ.

### DIFF
--- a/README.md
+++ b/README.md
@@ -93,14 +93,15 @@ supervised {
 
 ```scala mdoc:compile-only
 def computationR: Int = ???
-retry(RetryConfig.backoff(3, 100.millis, 5.minutes, Jitter.Equal))(computationR)
+retry(Schedule.exponentialBackoff(100.millis).maxRepeats(5)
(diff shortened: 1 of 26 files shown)


adamw:
@miciek if you'd have a spare minute, take a look at the `Schedule` class, maybe you'd have some comments/suggestions, as this tries to implement what you've talked about in your Scalar talk :)

miciek:
> @miciek if you'd have a spare minute, take a look at the `Schedule` class, maybe you'd have some comments/suggestions, as this tries to implement what you've talked about in your Scalar talk :)

@adamw Thanks for doing this! It looks great :) I probably wouldn't do it your way, but it's much better. I love the flexibility. I will play around with this one, thanks again!

I haven't run it yet but just looking at the code, I can't figure out what would be the preferred way of doing custom Schedules using your own `LazyList` operators? That's definitely the first thing I am trying out when I sit down to play with this version :)

PR #445, opened 2026-05-25
Add Scala Native support with ported Jox channels

adamw:
## Summary

- Add Scala Native 0.5.12 cross-compilation support to the `core` module
- Port the Jox channels library (lock-free segmented channel with select) from Java to Scala, using `AtomicLong`/`AtomicReference`/`AtomicReferenceArray` in place of `VarHandle` (not available in Scala Native)
- Native channel wrapper delegates to the Scala port in the same pattern as the JVM wrapper delegates to Java Jox

## Structure

- `core/shared/` — all platform-independent Ox code (structured concurrency, flows, resilience, scheduling)
- `core/jvm/` — JVM wrapper delegating to `com.softwaremill.jox` (unchanged behavior)
- `core/native/` — Scala Jox port (`ox.channels.jox` package) + native wrapper + tests

## Jox port details

Ported from jox 1.1.2 (`v1.1.2-channels` tag). Divergences documented in `core/native/src/main/scala/ox/channels/jox/README.md`:

1. `AtomicXxx` instead of `VarHandle` (Scala Native stubs VarHandle but doesn't implement it)
2. `AtomicReference[Segment]` parameter instead of `VarHandle` + owner object in `findAndMoveForward`
3. No `forEach`/`toList` on Source (provided by Ox wrapper)
4. No `Sink.trySend(value, channels...)` (provided by Ox select API)
5. Simplified `toString`

## Test plan

- [x] All 899 JVM tests pass (no regressions)
- [x] 51 native tests pass (rendezvous, buffered, unlimited, closed, trySend/tryReceive, interruption, select)
- [x] Native test compilation and linking verified with clang 19
- [ ] CI with Scala Native toolchain (requires clang in CI environment)


He-Pin:
AtomicXxx instead of VarHandle (Scala Native stubs VarHandle but doesn't implement it)
Is it possible to submit a pr to scala native,and implement VarHandle in scala native?

adamw:
@He-Pin as for `VarHandle` availability, maybe @WojciechMazur has some insights, as it's all his doing :)

WojciechMazur:
Regarding VarHandles - to be able to support them we'd need to introduce special handling in the compiler plugin - we don't have real reflection, so all values need to be resolved and preserved at compile time. It's doable, but was always delayed. In SN to replace VarHandles we typically use a combination of `scalanative.runtime.Intrinsics.classFieldRawPtr` + C stdlib atomics ([example here](https://github.com/scala-native/scala-native/blob/d12c9d9ea5f67d602d5db09e093776bb656f8ee8/javalib/src/main/scala/java/util/concurrent/ForkJoinPool.scala#L53-L86), and [another example for AtomicFieldUpdater](https://github.com/scala-native/scala-native/blob/main/scalalib/overrides-3/scala/collection/concurrent/TrieMap.scala.patch)) - if you want to reduce allocations that might be the best way go without waiting for changes in the upstream 
I can try to work on that for 0.5.13 when I find some spare time, but still most of the differences would be around the some additional allocation + indirection. Eventually it might be useful, but for first iteration j.u.c.atomic might be just fine

adamw:
@WojciechMazur great, yeah definitely using atomics is fine for the first pass

adamw:
Unfortunately there seems to be a problem with scala-native's virtual thread scheduling (as demonstrated by `NativeVirtualThreadScalabilityIssue.scala`), so we'll have to wait with merging until this is fixed.

PR #448, opened 2026-05-29
Fix AdaptiveRetry stopping on Left when shouldPayFailureCost returns false

### LINKED ISSUE #442: ox AdaptiveRetry stops retrying on Left when shouldPayFailureCost is false
## Project
ox (com.softwaremill.ox)

## Category
logic-error

## Severity
logic-error (retry incorrectly stops instead of continuing for free)

## Versions
All ox versions with AdaptiveRetry

## Reproducer

```scala
// In AdaptiveRetry.retryWithErrorMode, the afterAttempt callback:

// Left (error) case:
case Left(value) =>
  if config.resultPolicy.isWorthRetrying(value) then
    if shouldPayFailureCost(Left(value)) then ScheduleStop(!tokenBucket.tryAcquire(failureCost))
    else ScheduleStop.Yes  // BUG: stops retrying instead of continuing for free
  else ScheduleStop.Yes

// Right (non-success) case:
case Right(value) =>
  // ... (success check omitted)
  else if shouldPayFailureCost(Right(value)) then ScheduleStop(!tokenBucket.tryAcquire(failureCost))
  else ScheduleStop.No  // CORRECT: continues retrying for free
```

To trigger: Call `retryWithErrorMode` or `retryEither` with a `shouldPayFailureCost` that returns `false` for a `Left` error. The retry stops immediately on the first such error, even though:
1. `isWorthRetrying` returned `true` (error is retriable)
2. We decided not to pay the cost (free retry intended)

## Expected behavior
When `shouldPayFailureCost` returns `false`, the retry should continue without consuming tokens — matching the `Right` case behavior (`ScheduleStop.No`).

## Actual behavior
`ScheduleStop.Yes` (stop = true) is returned, terminating the retry immediately. This contradicts the documentation: "Penalty is paid only if it is decided to retry operation."

## Root cause
`core/src/main/scala/ox/resilience/AdaptiveRetry.scala`, line 79:
```scala
else ScheduleStop.Yes  // should be ScheduleStop.No
```

## Fix
Change line 79 from:
```scala
else ScheduleStop.Yes
```
to:
```scala
else ScheduleStop.No
```

## Duplicate check
- Searched ox GitHub issues for "shouldPayFailureCost", "AdaptiveRetry", "ScheduleStop" — no results related to this specific logic error

## Found by
Lane 2, Iteration 6, Strategy B3 (Error Handling Review)

## Verified: YES
**Evidence:** Lines 75-88 of AdaptiveRetry.scala:
```scala
case Left(value) =>
  if config.resultPolicy.isWorthRetrying(value) then
    if shouldPayFailureCost(Left(value)) then ScheduleStop(!tokenBucket.tryAcquire(failureCost))
    else ScheduleStop.Yes     // <-- line 79: STOPS retrying
  else ScheduleStop.Yes
case Right(value) =>
  ...
  else if shouldPayFailureCost(Right(value)) then ScheduleStop(!tokenBucket.tryAcquire(failureCost))
  else ScheduleStop.No        // <-- line 88: CONTINUES retrying
```
Confirmed: asymmetry between Left and Right cases. When `shouldPayFailureCost` returns `false`, the `Left` case returns `ScheduleStop.Yes` (stop) while the `Right` case returns `ScheduleStop.No` (continue). The `Left` case should also return `ScheduleStop.No` to continue for free.

## Minimal runnable reproducer

Save as `repro.scala` and run `scala-cli run --server=false repro.scala`:

```scala
//> using scala 3.6.4
//> using dep com.softwaremill.ox::core:1.0.4

// Bug 020: ox AdaptiveRetry stops retrying on Left when shouldPayFailureCost returns false.
//
// In AdaptiveRetry.retryWithErrorMode the `Left` branch returns ScheduleStop.Yes when
// `shouldPayFailureCost(Left(_))` is false — incorrectly terminating retry. The symmetric
// `Right` branch returns ScheduleStop.No (continues for free). Per the docs:
//   "Penalty is paid only if it is decided to retry operation."
// So when we explicitly say "don't pay", we should still retry — for free.

import ox.resilience.*
import ox.scheduling.Schedule

@main def repro(): Unit =
  val ar = AdaptiveRetry.default // TokenBucket(500), failureCost=5, successReward=1

  // Allow up to 5 attempts (1 initial + 4 retries) with no delay.
  val schedule = Schedule.immediate.maxAttempts(5)

  // Count how many times the operation actually runs.
  var attempts = 0
  val op: () => Either[String, Int] = () =>
    attempts += 1
    Left(s"transient-error-$attempts")

  // shouldPayFailureCost = false  =>  we explicitly opt out of paying the cost.
  // Expected: retry continues for free (5 attempts), like the Right branch does.
  // Actual (bug): retry stops after the first Left (1 attempt).
  val result: Either[String, Int] =
    ar.retryEither(schedule, (_: Either[String, Int]) => false)(op())

  val expectedAttempts = 5
  val actualAttempts = attempts

  println(s"shouldPayFailureCost = (_ => false)")
  println(s"schedule = Schedule.immediate.maxAttempts(5)")
  println(s"result = $result")
  println(s"expected attempts = $expectedAttempts (free retries, like the Right branch)")
  println(s"actual   attempts = $actualAttempts")
  if actualAttempts < expectedAttempts then
    println(s"BUG REPRODUCED: AdaptiveRetry stopped on Left after $actualAttempts attempt(s) " +
            s"even though shouldPayFailureCost returned false (free retry was intended).")
  else
    println("No bug observed.")
```

---

Verified locally:
1. cross-vendor adversarial gate (codex prover + claude skeptic + 2 independent judges, distinct `judge_id` by vendor) returned `accepted`
2. `scala-cli run --server=false` on the repro above reproduces the symptom (exit code / output mismatch)

AI-assisted report. If I missed context or it's intended behavior, sorry — happy to close. I batch-audited my older filings yesterday and self-closed 8 false positives, so I'm trying to be conservative now.

adamw:
Fixes #442.

When AdaptiveRetry encounters a Left error and shouldPayFailureCost returns false, retrying should continue. Currently, the retry stops in this scenario.

This commit adds a failing test that demonstrates the expected behavior: retries should continue for the configured schedule even when the cost is not paid (shouldPayFailureCost returns false).

Closes softwaremill/ox#442.

### DIFF
--- a/core/src/test/scala/ox/resilience/ImmediateRetryTest.scala
+++ b/core/src/test/scala/ox/resilience/ImmediateRetryTest.scala
@@ -210,4 +210,21 @@ class ImmediateRetryTest extends AnyFlatSpec with EitherValues with TryValues wi
     result.value shouldBe message
     counter shouldBe 6
 
+  it should "not pay exceptionCost if Left error is going to be retried and shouldPayPenaltyCost returns false" in:
+    // given
+    var counter = 0
+    val errorMessage = "boom"
+
+    def f: Either[String, Int] =
+      counter += 1
+      Left(errorMessage)
+
+    val adaptive = AdaptiveRetry(TokenBucket(2), 1, 1)
+    // when
+    val result = adaptive.retryEither(Schedule.immediate.maxRetries(5), (_: Either[String, Int]) => false)(f)
+
+    // then
+    result.left.value shouldBe errorMessage
+    counter shouldBe 6
+
 end ImmediateRetryTest


adamw:
The failure is clear. Here's the summary:

---

**CI failure: `ImmediateRetryTest` — one test, both Java 21 and 24**

```
- should not pay exceptionCost if Left error is going to be retried
  and shouldPayPenaltyCost returns false *** FAILED ***
  1 was not equal to 6 (ImmediateRetryTest.scala:228)
```

The test asserts that when `shouldPayPenaltyCost` returns `false` and the result is a `Left` that will be retried, the penalty cost counter stays at 6 (untouched across retries). Instead it is 1, meaning the implementation is charging the penalty cost on the first attempt and stopping, rather than skipping the cost and continuing to retry. This is the intentional failing test added to document the bug this PR is meant to fix.

PR #455, opened 2026-06-07
Make Flow.range support descending ranges with negative step

### LINKED ISSUE #439: Flow.range with a negative step emits only the first element
`Flow.range(5, 1, -1)` emits `[5]` instead of `[5, 4, 3, 2, 1]`. The continuation condition only handles ascending ranges.

Reproducer (scala-cli):

```scala
//> using scala 3.8.3
//> using dep com.softwaremill.ox::core:1.0.4

import ox.flow.Flow

@main def repro(): Unit =
  println(Flow.range(5, 1, -1).runToList())   // List(5)            — expected List(5, 4, 3, 2, 1)
  println(Flow.range(10, 0, -2).runToList())  // List(10)           — expected List(10, 8, 6, 4, 2, 0)
```

Source:

https://github.com/softwaremill/ox/blob/770bed97a247a2536832ff7485953d75f53ccbeb/core/src/main/scala/ox/flow/FlowCompanionOps.scala

```scala
def range(from: Int, to: Int, step: Int): Flow[Int] = usingEmitInline: emit =>
  var t = from
  repeatWhile:
    emit(t)
    t = t + step
    t <= to                       // wrong direction when step < 0
```

For descending ranges (`step < 0`), the loop guard should be `t >= to`. Scala's own `5 to 1 by -1` follows that convention.

There's also a corner case in the other direction: `Flow.range(1, 5, -1)` would loop until Int underflow, since `t <= to` stays true. Probably worth a `step != 0` precondition while in the area.

Suggested fix:

```scala
def range(from: Int, to: Int, step: Int): Flow[Int] =
  require(step != 0, "step must be non-zero")
  usingEmitInline: emit =>
    var t = from
    repeatWhile:
      emit(t)
      t = t + step
      if step > 0 then t <= to else t >= to
```

Happy to PR.

adamw:
Fixes #439 — Flow.range with a negative step was only emitting the first element due to incorrect termination logic.

The original implementation used a fixed termination condition (`t <= to`) that didn't account for negative steps. With a negative step, this condition would cause the loop to terminate prematurely.

The fix updates the range method to:
- Require non-zero step values
- Use conditional termination logic: `t <= to` for positive steps, `t >= to` for negative steps
- Document the descending range behavior with examples

Test cases for descending ranges have been added to verify the fix works correctly.

Closes softwaremill/ox#439.

### DIFF
--- a/core/src/main/scala/ox/flow/FlowCompanionOps.scala
+++ b/core/src/main/scala/ox/flow/FlowCompanionOps.scala
@@ -96,6 +96,7 @@ trait FlowCompanionOps:
         emit(t)
         t = t + step
         if step >= 0 then t <= to else t >= to
+  end range
 
   /** Creates a flow which emits the first element of tuples returned by repeated applications of `f`. The `initial` state is used for the
     * first application, and then the state is updated with the second element of the tuple. Emission stops when `f` returns `None`,
--- a/core/src/main/scala/ox/flow/FlowCompanionOps.scala
+++ b/core/src/main/scala/ox/flow/FlowCompanionOps.scala
@@ -85,13 +85,17 @@ trait FlowCompanionOps:
       emit(t)
       t = f(t)
 
-  /** Creates a flow which emits a range of numbers, from `from`, to `to` (inclusive), stepped by `step`. */
-  def range(from: Int, to: Int, step: Int): Flow[Int] = usingEmitInline: emit =>
-    var t = from
-    repeatWhile:
-      emit(t)
-      t = t + step
-      t <= to
+  /** Creates a flow which emits a range of numbers, from `from`, to `to` (inclusive), stepped by `step`. A negative `step` produces a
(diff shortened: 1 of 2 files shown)


adamw:
The failure is clear. Here's the summary:

---

**CI failure: `Flow.range` negative step not yet implemented**

Both `ci (21)` and `ci (24)` fail on the same test in `FlowOpsFactoryMethodsTest`:

```
- should produce a descending range with a negative step *** FAILED ***
  List(5) was not equal to List(5, 4, 3, 2, 1) (FlowOpsFactoryMethodsTest.scala:30)
```

`Flow.range(5, 0, -1)` (or equivalent) currently returns only `List(5)` instead of `List(5, 4, 3, 2, 1)`. The termination condition in `Flow.range` does not account for a negative step — it stops immediately after the first element because it checks `current < end` (or similar ascending-only guard) even when stepping downward.

PR #480, opened 2026-07-20
Add resource-only scopes (resourceScope / ResourceScope)

adamw:
Adds **resource scopes**: scopes which allow attaching resources (`useInScope` / `releaseAfterScope` style), but are not full concurrency scopes — no forking, no threads started while the body runs. Every concurrency scope can be used where a resource scope is expected (via subtyping: `OxUnsupervised extends ResourceScope`). This is ox's analogue of `scala.util.Using.Manager`, and lets methods declare exactly the capability they need (`using ResourceScope` — attach cleanup, without claiming the ability to fork).

Design highlights (adversarially critiqued before implementation):

- **Compile-time guard**: `resourceScope` cannot be started where a concurrency scope is lexically visible (via a `NotGiven`-based given with a custom error message) — forks started within a lexically visible resource scope could outlive it. The recommended structure is extracting the scope to a capability-free method.
- **Finalizer-list freeze**: registering a resource after its scope ended (possible only via explicitly-leaked capabilities) now throws `IllegalStateException` instead of being silently lost; `useInScope` releases the just-acquired resource in that case, so cleanup is never lost. Applies uniformly to concurrency scopes.
- **Binary compatibility preserved**: the four resource functions now take `using ResourceScope`; bridges with the old `using OxUnsupervised` JVM signatures are kept via `@targetName` (with distinct Scala-level names, avoiding overload ambiguity) — MiMa passes with no new filters. Source compatibility is unaffected.
- Nesting/`ForkLocal` semantics pinned by tests: resources attach to the nearest enclosing scope; resource-scope finalizers see the fork-local values of the scope's own level.

### DIFF
--- a/core/src/main/scala/ox/Ox.scala
+++ b/core/src/main/scala/ox/Ox.scala
@@ -8,7 +8,7 @@ import ox.internal.ThreadHerd
 import java.util.concurrent.atomic.AtomicReference
 import scala.annotation.implicitNotFound
 
-/** Capability granted by a [[resourceScope]], as well as, via subtyping, by any concurrency scope ([[supervised]], [[supervisedError]],
+/** Capability granted by a [[resourceScope]] and, via subtyping, by any concurrency scope ([[supervised]], [[supervisedError]],
   * [[unsupervised]]).
   *
   * Represents a capability to register resources (e.g. using [[useInScope]] or [[releaseAfterScope]]) to be released when the scope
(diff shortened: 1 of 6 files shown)


adamw:
Release-note-worthy behavior change (surfaced in review): a finalizer which registers another finalizer on the same scope previously had the registration **silently lost**; it now fails the scope with an `IllegalStateException` (suppressed into the scope's result). No correct program could depend on the old behavior — the registered finalizer never ran.

PR #482, opened 2026-07-21
Add computeIntensive for running CPU-intensive operations off virtual threads

adamw:
## Motivation

Virtual threads are not preempted by the JVM scheduler: a long-running, CPU-bound computation monopolizes a carrier thread, and a handful of such computations can starve all other virtual threads in the process (JEP 444 explicitly excludes time-sharing). `cede()` (#481) covers short, instrumentable bursts; this PR adds the complement for long-running or non-instrumentable compute, following the pattern the JDK team recommends (run such work on platform threads, letting the OS preempt) and the shape other runtimes converged on (Kotlin's `Dispatchers.Default`, Tokio's dedicated compute pool, Reactor's `Schedulers.parallel()`).

## Changes

* `computeIntensive(f)` / `computeIntensive(executor)(f)`: runs `f` on a pool of platform threads, blocking the calling (virtual) thread until completion — structured by construction (the computation never outlives the enclosing scope), composes with `fork`/`par`/`race`/`timeout`.
* Default pool: lazily-created, global, fixed size = available processors, daemon platform threads (`ox-compute-N`); replaceable via `setOxComputeExecutor` (mirroring `setOxThreadFactory`) or `OxApp.Settings.computeExecutor` (with binary-compat overloads).
* Cancellation: interrupting the caller interrupts the pool task and then **awaits its completion** (strict structure); a not-yet-started task never runs; a task-thrown `InterruptedException` is correctly distinguished from caller interruption; the interrupt never leaks to a subsequent task on the reused pool thread.
* Nesting: nested calls targeting the same executor (by reference) run inline, avoiding fixed-pool deadlock; different executors submit normally.
* Scope context is deliberately not propagated: `ForkLocal`s read defaults, `fork` inside the computation fails fast.
* Docs: new `doc/utils/compute-intensive.md` page + cross-links from the `cede()` docs.
* 16 new tests, including interrupt-then-await semantics, cancel-before-start, nesting, `either` interop, and a time-limited starvation smoke test.

## Notes

* The positive `setOxComputeExecutor` / `OxApp.Settings.computeExecutor` wiring path is deliberately untested: it mutates JVM-global state and would poison other suites (same situation as the existing `threadFactory` setting, which is also untested). Only the "throws after first use" path is covered.
* Full core suite: 950 passed, 0 failed.


adamw:
Superseded by #483 (recreated due to a GitHub-side issue with merging this PR).

